PR #1, opened 2021-06-18
Generative rotated molecules from a finite set of particules

NicolasLegendre1:
The program simulates a uniform set of rotations for a molecule in 3d considered as a finite set of particules


sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+    volume : ndarray volume of the molecule.
+    SIZE : int size of the volume.
+    LONG : int center the molecule around zero.
+
+    Returns
+    -------
+    volume : ndarray volume of the molecule.
+
+    """
+    for _, point in enumerate(points.T):
```
ninamiolane, line 60: `for point in points.T` should work

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+            for j in range(SIZE):
+                for k in range(SIZE):
+                    volume[i][j][k] += np.exp(-dist([i/LONG-SIZE/LONG/2,
+                                                     j/LONG-SIZE/LONG/2,
+                                                     k/LONG-SIZE/LONG/2],
+                                                    point)/2)
+    return volume
+
+
+def simulate_3D_volumes(particules, n_volumes, img_size, long=2):
```
ninamiolane, line 71: `simulate_volumes` should be enough (trying to be short here)

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+import torch
+import os
+
+CUDA = torch.cuda.is_available()
+
+if CUDA:
+    Main_dir = "/scratch/ex-kdd-1/NicolasLegendre/Cryo/Data"
+else:
+    Main_dir = os.getcwd()+"\\Data\\"
```
ninamiolane, line 18: use `black` on your code

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+    point2 : tuple float.
+
+    Returns
+    -------
+    distance between point1 and point2.
+
+    """
+    a, b, c = point1
+    d, e, f = point2
+    return (a-d)**2+(b-e)**2+(c-f)**2
```
ninamiolane, line 42: how about: `np.linalg.norm(point1-point2)`? no need to reimplement the euclidean dist in a new function IMO

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+    SIZE : int size of the volume.
+    LONG : int center the molecule around zero.
+
+    Returns
+    -------
+    volume : ndarray volume of the molecule.
+
+    """
+    for _, point in enumerate(points.T):
+        for i in range(SIZE):
```
ninamiolane, line 61: There should be a way to avoid this triple for loop

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+    -------
+    distance between point1 and point2.
+
+    """
+    a, b, c = point1
+    d, e, f = point2
+    return (a-d)**2+(b-e)**2+(c-f)**2
+
+
+def modif_weight(points, volume, SIZE, LONG):
```
ninamiolane, line 45: lowercase letters 🙏

sim_3d_volumes.py
ninamiolane: btw, I thought that `points` was a list, according to your docstring, can you transpose a list??

We need unittests to make sure this works

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+    long : float, optional center the molecule around zero.
+
+    Returns
+    -------
+    volumes : ndarray 3D map of the molecule.
+    qs : dataframe describes rotations in quaternions.
+
+    """
+    rots, qs = coords.uniform_rotations(n_volumes)
+    volumes = np.zeros((n_volumes, img_size, img_size,
```
ninamiolane, line 88: `(n_volumes,) + (img_size,) * 3` should work

sim_3d_volumes.py
ninamiolane: maybe `modify_weights`? it's one more letter, and at least we have the full verb

sim_3d_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+                       img_size))
+    for idx in range(n_volumes):
+        if idx % (n_volumes/10) == 0:
+            print(idx)
+        points = rots[idx].dot(particules)
+        volumes[idx] = modif_weight(points, volumes[idx], img_size, long)
+    qs = [np.array2string(q) for q in qs]
+    qs = pd.DataFrame(qs)
```
ninamiolane, line 96: we need to decide which format we want to use across the `compSPI` repo... do we want to enforce `np.array` or `pd.DataFrame` @fredericpoitevin ?

It could become confusing if some functions return mostly pd.dataframes and other work with numpy...

ninamiolane (changes requested):
Very nice, @NicolasLegendre1 , great contributions throughout all the repos!! 🎉 💯 

The comments from your other PR hold for this too:
- IMPORTANT: add unit-tests
- docstring syntax is not correct
- the use of uppercase letters is not correct: if you start a variable with a uppercase letter, python thinks it's a constant or a Python class.

sim_3d_volumes.py
fredericpoitevin: Ah yes, I agree we need to enforce one. I'd vote in favor or `np.array` everywhere. I think not adding `pandas` to the environments (TO DO) would be a pretty straightforward way to enforce this. 
I hope to push soon some information about conda environments, containers and more generally how to install and use the different repos.

geoffwoollard: I vote numpy > pandas.

I'm a big pandas fan because there are convenient things in pandas.DataFrame and pandas.Series like .plot, .value_counts and .melt & .pivot a la http://hadley.nz/

However, I have a hunch that there could be performance issues if things are done with pandas vs numpy. So my vote is numpy > pandas.

NicolasLegendre1: so do we all converge to np.array format?

sim_3d_volumes.py
NicolasLegendre1: I don't see exactly how to avoid that

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
+""" Generate 3D map of molecules
+"""
```
fredericpoitevin, line 2: To follow the style guideline closely this should end with a period and be only one line.

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+
+def modify_weight(points, volume, size, long):
+    """
+    Fill an empty volume with particles.
+
+    Parameters
+    ----------
+    points : array, list particules position.
+    volume : ndarray, volume of the molecule.
+    SIZE : int, size of the volume.
```
fredericpoitevin, line 31: the parameters are lower case

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+    LONG : int, center the molecule around zero.
+
+    Returns
+    -------
+    volume : ndarray, volume of the molecule.
+
```
fredericpoitevin, line 37: NIT: add an Example section, in addition to Parameters and Returns.

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+Particules = np.asarray([[1, 0, 0, 3], [0, 8, 0, 5], [0, 0, 7, 9]])
+N_particules = 2000
+Vol_size = 64
+Name = '4_points_3D'
+Long = 2
+
+
+def modify_weight(points, volume, size, long):
+    """
+    Fill an empty volume with particles.
```
fredericpoitevin, line 25: It sounds like the volume does not really have to be empty? 
Maybe this is more accurate: "Update a volume with new particles"

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    volumes, labels = simulate_volumes(
+        particules, n_volumes, vol_size, long)
+    np.save(main_dir + name + '_molecules.npy', volumes)
+    np.save(main_dir + name + '_labels.npy', volumes)
+
+    return volumes, labels
+
+
+A = save_volume(Particules, N_particules, Vol_size, Main_dir, Name, long=Long)
```
fredericpoitevin, line 103: To make this executable as a script or interactively while also allowing to use the functions in a module, it is recommended to define a function `main()` and add this at the end of the file:
```
def main():
  A = save_volume(Particules, N_particules, Vol_size, Main_dir, Name, long=Long)

if __name__ == "__main__":
  # execute only if run as a script
  main()
```

fredericpoitevin: see e.g. https://docs.python.org/3/library/__main__.html

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+if CUDA:
+    Main_dir = "/scratch/ex-kdd-1/NicolasLegendre/Cryo/Data"
+else:
+    Main_dir = os.getcwd() + "\\Data\\"
+
+Particules = np.asarray([[1, 0, 0, 3], [0, 8, 0, 5], [0, 0, 7, 9]])
+N_particules = 2000
+Vol_size = 64
+Name = '4_points_3D'
+Long = 2
```
fredericpoitevin, line 20: I would not hard code those parameters here. Please see comment below and make these arguments default of `main()`

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
...
+import torch
+import coords
+
+
+CUDA = torch.cuda.is_available()
+
+if CUDA:
+    Main_dir = "/scratch/ex-kdd-1/NicolasLegendre/Cryo/Data"
+else:
+    Main_dir = os.getcwd() + "\\Data\\"
```
fredericpoitevin, line 14: It sounds like we could make this more general by taking these lines out as I do not see anywhere else a `torch` or `CUDA` dependency. This type of check could be done in an external script.

sim_3d_volumes.py
```diff
@@ -0,0 +1,103 @@
+""" Generate 3D map of molecules
+"""
+import os
+import numpy as np
+import torch
+import coords
```
fredericpoitevin, line 6: These modules are not used, I would remove those lines.

fredericpoitevin (changes requested):
Great job @NicolasLegendre1 ! I hope my comments are not too redundant with Nina's!
Do not hesitate to reach out if you want to discuss them in person.

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+    Returns
+    -------
+    None.
+
+    """
+    _ = save_volume(particules, n_particules, vol_size,
+                    main_dir, name, long)
+
+
+# execute only if run as a script.
```
ninamiolane, line 125: I dont think that we need this comment

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+
+def save_volume(particules, n_volumes, vol_size, main_dir, name, long=2):
+    """
+    Save the simulate volumes and meta_data.
+
+    Parameters
+    ----------
+    particules : array, position of particules.
+    n_particules : int, number of data.
+    vol_size : int, size of the volume.
```
ninamiolane, line 84: is `vol_size` different from `size` and `img_size`? if not, we should use the same name. (to avoid confusing users too much)

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def save_volume(particules, n_volumes, vol_size, main_dir, name, long=2):
+    """
+    Save the simulate volumes and meta_data.
+
+    Parameters
+    ----------
+    particules : array, position of particules.
+    n_particules : int, number of data.
```
ninamiolane, line 83: this should be `n_volumes` from the signature of the function

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+    return volume
+
+
+def simulate_volumes(particules, n_volumes, img_size, long=2):
+    """
+    Update a volume with new particles.
+
+    Parameters
+    ----------
+    particules : ndarray, volume of the molecule.
```
ninamiolane, line 54: depending on the functions, `particules` seems to refer either to a volume (here) or to positions of particles (below): use two different names?

ninamiolane (changes requested):
Great!! 🎉 Thank you for addressing the first comments!

Main remarks:
- We still need some unit-tests
- It would be great if you could follow Python standards in naming variables: no uppercase letters when it is not a Class or a CONSTANT, etc =D 
- The API of the functions needs some cleaning: some parameter names are not very intuitive (`long` refers to the center of the volume?), some parameters have different names to refer to the same concept (`img_size` and `vol_size`?), or have the same name to refer to different concepts (eg `particles`?). See my more detailed comments in the PR. Or maybe the docstrings misled me? In any case, the API is what our users see first. If they need to use different parameters names depending on the function they use, it will be confusing for them. Therefore, the API should be as logical, intuitive, consistent and rigorous as possible.

Great job though, the PRs will be faster once we get the good habits in!!

ninamiolane:
The file could be named `sim_volumes.py`, since 3D volumes is probably redundant?

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+CUDA = torch.cuda.is_available()
+
+if CUDA:
+    Main_dir = "/scratch/ex-kdd-1/NicolasLegendre/Cryo/Data"
+else:
+    Main_dir = os.getcwd() + "\\Data\\"
+
+Particules = np.asarray([[1, 0, 0, 3], [0, 8, 0, 5], [0, 0, 7, 9]])
```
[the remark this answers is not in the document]
NicolasLegendre1: is that not a constant?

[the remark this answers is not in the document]
ninamiolane: Right, sorry, it is a constant. Constants need to be all uppercase letters: so PARTICULES here :D

sim_3d_volumes.py
```diff
@@ -0,0 +1,127 @@
...
+N_particules = 2000
```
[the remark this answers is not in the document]
ninamiolane: Same here

sim_3d_volumes.py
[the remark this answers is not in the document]
NicolasLegendre1: okay done

PR #7, opened 2021-10-26
Feature: wrapper function definitions

jedyeo:
We've defined what we'd like to achieve with this wrapper and have written docstrings for each function. Please review and let us know!


src/temp_workspace/input/sim_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    optics_defocusout: None     # file to write defocus values
+
+  detector_parameters:
+    detector_Nx: 5760           # number of pixels along X axis
+    detector_Ny: 4092           # number of pixels along Y axis
+    detector_pixel_size: 5      # [um]
+    detector_gain: 2            # average number of counts per electron
+    detector_Q_efficiency: 0.5  # detector quantum efficiency
+    noise: 'no'                 # whether quantized electron waves result in noise
+    MTF_params: [0,0,1,0,0]     # to be described. [0,0,1,0,0] is perfect detector.
```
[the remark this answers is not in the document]
ninamiolane: @geoffwoollard I think I agree with this:
- if the community uses the star format then we can use the star format

My only concern in moving from yml to star would be: 
- is the star format as easily readable as is the .yml format? 

The advantages with the yml are that:
- everybody knows it (e.g. machine learners who are not necessarily used to cryoEM conventions)
- it is easily readable on GitHub, e.g. in this PR it is easy to see how parameters are set up.

If the star format is easily readable, then we can move to it. Otherwise, I'm not sure...

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+    ----------
+    path_config : str
+        Relative path to YAML file containing file paths for TEM Simulator.
+    sim_config : str
+        Relative path to YAML file containing simulator paths for TEM Simulator.
+
+    """
+
+    def __init__(self, path_config, sim_config):
+        self.file_paths = self._getConfigFromYaml(path_config)
```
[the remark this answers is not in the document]
ninamiolane: And maybe this should be called `path_dict` for consistency in the namings:
- sim -> sim_config -> sim_dict
- path -> path_config -> path_dict

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+            Relative path to YAML file containing parameters for TEM Simulator
+        Returns
+        -------
+        classified_params : dict
+            Dictionary containing grouped parameters for TEM Simulator, with keys:
+                seed : str maps to int
+                    Seed for TEM Simulator
+                particle_mrcout : str maps to bool
+                    Flag for optional volume map of sample
+                sample_dimensions : str maps to
```
ninamiolane, line 60: str maps to ... ?

ninamiolane: should` sample_dimensions` be called `specimen_grid_params`? it would be more explanatory and consistent with the other namings

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+        file_paths = {}
+        return file_paths
+
+    @staticmethod
+    def create_crd_file(file_paths, sim_param_arrays, pad):
+        """Format and write molecular model data to crd_file for use in TEM-simulator.
+
+        Parameters
+        ----------
+        file_paths : dict of type str to str
```
ninamiolane, line 129: is this the same as `self.file_paths` from the constructor? if so, can we access it with `self.file_paths` directly?
same question for `sim_param_arrays` (which, in this case, should be called `self.sim_dict`?)

src/transmission_electron.py
ninamiolane: `specimen_grid_params` is also the name that appears in `sim_config.yml`.

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+                    List containing the detector parameters
+                optics_params : str maps to list
+                    List containing the optic parameters
+
+        """
+        return None
+
+    @staticmethod
+    def get_image_data(file_paths, sim_path):
+        """Run simulator and return data.
```
ninamiolane, line 149: How is this function different from `run` then?

ninamiolane: file_paths -> self.path_dict
sim_path -> self.sim_dict?

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+        # inp_file = Path(file_paths["inp_file"])
+
+        # cmd = "{0} {1}".format(SIMULATOR_BIN, inp_file)
+        # os.system(shlex.quote(cmd))
+
+        return []
+
+    @staticmethod
+    def generate_parameters_dictionary(file_paths, sim_params):
+        """Compile experiment data into .inp friendly file for use in TEM-simulator.
```
ninamiolane, line 176: Can you give a brief explanation of what is the `.inp` format (1 line)

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Parameters
+        ----------
+        file_paths : dict of type str to str
```
ninamiolane, line 180: self.path_dict?

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+             Dict of file paths that includes keys:
+                pdb_file
+                    Relative path to pdb input file
+                crd_file
+                    Relative path to populated crd file
+                mrc_file
+                    Relative path to desired output mrc file
+                log_file
+                    Relative path to desired output log file
+        sim_params : dict
```
ninamiolane, line 190: self.sim_dict?

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+        file_paths : dict of type str to str
+            Dict of file paths that includes keys:
+                inp_file
+                    Relative path to input file to be populated with parameters
+        """
+        return None
+
+    @staticmethod
+    def extract_particles(micrograph, sim_param_arrays, file_paths, pad):
+        """Format and write molecular model data to crd_file for use in TEM-simulator.
```
ninamiolane, line 224: This docstring is not correct (this is the same docstring as another function above)

src/transmission_electron.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Parameters
+        ----------
+        micrograph : arr
+            Array containing TEM-simulator micrograph output
+        sim_param_arrays : dict
```
ninamiolane, line 230: self.sim_dict?

ninamiolane (changes requested):

ninamiolane:
@geoffwoollard @jedyeo Sorry for the delay in sending this review!

This is really nice, well-written! My main comments are:
- the organization of the files in folders: 
   - it seems that some folders create an unnecessary arborescence in our files: why `src`?
   - others have non-explanatory names that could become misleading on the long run: what is `temp_workspace`
- the use of the dictionaries of parameters (from paths or from sim) is not consistent:
  - the dictionnaries of parameters are stored as attributes of the TEMSimulator object, which is great!
  - once they are stored, then can simply be accessed with e.g. self.sim_dict
  - thus, I don't understand why these dictionaries are passed as inputs of many of the functions.
- let's be extra rigorous in variable namings (I have put some comments on that topic in the code review): it is not fun to think about, but believe me it will save us headaches in the future if our variables are well-named

But this is great, very clean otherwise, and the object oriented structure is definitely the way to go. Thank you 🎉

geoffwoollard:
>   * it seems that some folders create an unnecessary arborescence in our files: why `src`?

Good point @ninamiolane . 

@jedyeo @calhep @arjunsingh3600 @thisTyler: if you look at the simSPI repo (e.g. if you clone it), you'll have folders `simSPI/tests` and `simSPI/simSPI` and `simSPI/.github`. The folder `simSPI/simSPI` _is_ for source code, so there is not really a need to make a src dir.

There are already a few .py files and .yaml files for the temsim wrapper, so perhaps you could put temsim specific things in simSPI/simSPI/tem_simulator:

```
simSPI/simSPI/tem_simulator/helpers.py
simSPI/simSPI/tem_simulator/transmission_electron.py
simSPI/simSPI/tem_simulator/sim_config.yaml
simSPI/simSPI/tem_simulator/path_config.yaml
```

ninamiolane (approved):
Much better, thank you for this great clean up!

Happy to merge this now (have a look at my comment on the file arborescence) and see which code you put in which function!

For next PR, it would be great if you could introduce typing: https://docs.python.org/3/library/typing.html

Thanks!

geoffwoollard:
For the tests, you could write some minimal tests that check the type of the return object, or just pass, since there is nothing much to test yet!

The coverage comes from the percentage of lines that get ran in the function to be tested. For instance if there is an if statements and the test doesn't check all control flow paths, then it would not be 100%.

geoffwoollard (approved):

PR #9, opened 2021-10-28
Temsim container

fredericpoitevin:
Attempt at setting up CI for containerization, specifically a TEM-simulator container. See https://github.com/compSPI/simSPI/pull/7#issuecomment-953528223 for more details.

### DIFF
--- a/.github/workflows/temsimulator.yml
+++ b/.github/workflows/temsimulator.yml
@@ -1,0 +1,49 @@
+name: "Building TEM simulator container"
+
+on:
+  push:
+    branches: [ master, temsim-container ]
+
+  pull_request:
+    branch: [ master, temsim-container ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Check Out Repo
+        uses: actions/checkout@v2
+
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./
+          file: ./docker/TEM-simulator/Dockerfile
+          builder: ${{ steps.buildx.outputs.name }}
+          push: true
+          tags: ${{ secrets.DOCKER_HUB_USERNAME }}/simSPI:latest
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
+
+    - name: Image digest
+      run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/temsimulator.yml
+++ b/.github/workflows/temsimulator.yml
@@ -44,6 +44,5 @@ jobs:
           tags: ${{ secrets.DOCKER_HUB_USERNAME }}/simSPI:latest
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
-
-    - name: Image digest
-      run: echo ${{ steps.docker_build.outputs.digest }}
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}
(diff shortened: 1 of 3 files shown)


ninamiolane:
👮‍♀️ 😄  hehehe

geoffwoollard:
Ok I think I see from Nina's comments that this is doing some "installation stuff" on the cloud, and then making executable code available on the computer its running on.

Rather than installing tem simulator from scratch on whatever computer the repo is cloned to.

On what machines do the paths in docker/TEM-simulator/Dockerfile refer to? Like `/work`?

.github/workflows/temsimulator.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  push:
+    branches: [ master, temsim-container ]
+
+  pull_request:
+    branch: [ master, temsim-container ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
geoffwoollard, line 12: is this some cloud "container"?

.github/workflows/temsimulator.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
```
geoffwoollard, line 27: what and were is Docker Hub?

.github/workflows/temsimulator.yml
```diff
@@ -0,0 +1,48 @@
...
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./
+          file: ./docker/TEM-simulator/Dockerfile
+          builder: ${{ steps.buildx.outputs.name }}
+          push: true
+          tags: ${{ secrets.DOCKER_HUB_USERNAME }}/simspi-temsimulator:latest
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
```
geoffwoollard, line 46: does this define where the build software ends up locally?

fredericpoitevin:
> > Do we need to build and push to dockerhub everytime someone creates a PR? Maybe we can change the `on` field from the github workflow, so that a new image is built and push only if the dockerfile has been modified, so that it does not slow down everybody's PRs?
> 
> Good idea! I'll see how to do that.
> 
> > NIT: maybe the workflow could be called `tem_dockerbuild.yml` since it is not simulating anything but rather building the image?
> 
> Good point! Let's not merge until this is perfect :D

@ninamiolane - I believe this is better now.

.github/workflows/temsimulator.yml
fredericpoitevin: so here we are asking GitHub Actions to build our container as if we were on a Linux machine whose OS is Ubuntu.

.github/workflows/temsimulator.yml
fredericpoitevin: I gave my credentials so the image is being built and pushed here: https://hub.docker.com/layers/174436511/fpoitevi/simspi-temsimulator/latest/images/sha256-2b0366b0d6f6d764f99b336a50e9bbf2f64e707fe49769d0685672484882ed08?context=repo

.github/workflows/temsimulator.yml
fredericpoitevin: This is just use for "caching" - meaning that next time we run this workflow to build the same container, some of it will already be "cached" at this path, helping speed things up a bit.

fredericpoitevin:
> Ok I think I see from Nina's comments that this is doing some "installation stuff" on the cloud, and then making executable code available on the computer its running on.
> 
> Rather than installing tem simulator from scratch on whatever computer the repo is cloned to.
> 
> On what machines do the paths in docker/TEM-simulator/Dockerfile refer to? Like `/work`?

So `/work` is a path "in the container". It sounds very abstract right now, but with a little bit of practice things get much clearer.

At this point, we have "built" and "pushed" an "image" to the "cloud". For you to use it, the next steps are to "pull" it and "run" it as a container. If you have a Linux machine handy, you could install Singularity and do the following:
```bash
singularity pull docker://fpoitevi/simspi-temsimulator:latest
```
This would "pull" the Docker image from the cloud and convert it to a Singularity image: a `simspi-temsimulator_latest.sif` file. You are now equipped with an image that you can run as a container - what does this mean? It means you can "execute" the container and work in it as if you were accessing a different computer.

Say you do this in a Terminal window on your Linux machine:
```bash
singularity exec simspi-temsimulator_latest.sif /bin/bash
```
This will just appear as a regular Terminal bash line (starting with `Singularity>`) but now you can run TEM simulator since you are in the container where it is installed.

geoffwoollard:
I did some work on the tests, and merged into main https://github.com/compSPI/simSPI/pull/12

ninamiolane (approved):
Looking good, thank you and sorry for the delay in reviewing this!

As a side note, @fredericpoitevin , could you maybe add something in the README on how to use this image?

PR #15, opened 2021-11-14
TEM Simulator Wrapper Class

jedyeo:
This is the wrapper for the TEM simulator. 

Tests are currently WIP, and the code needs some style checking/cleaning up. That being said the core implementation is there and the output is looking good (particle stack). :)


jedyeo:
@calhep @arjunsingh3600 @thisTyler

ninamiolane:
Thank you very much, and congratulations for all of this code! This looks very promising!

Yet, it would be way easier and much faster to review this work if you send smaller blocks of code. This PR is very very big which makes it almost impossible to be reviewed properly at once. It is more efficient to send less code in very small PRs, that can be reviewed and merge in a day or 2, and we can iterate quickly.

Thus, could you close this PR and resend it as:
- several smaller PRs (muuuuuch less code per PR), where each PR has:
  - unit-tests
  - cleaner code (I see many left-over comments or camelCase that should not be there; also please use `logging` instead of prints)
  - documentation.

Thank you!

jedyeo:
Will do!

PR #16, opened 2021-11-15
Add sim and path config yaml files

jedyeo:
Add sim and path config yaml files for the simulator to use.

Code to receive and parse these yamls will be added in a future PR (soon).

@calhep @thisTyler @arjunsingh3600


temp_workspace/input/sim_config.yaml
```diff
@@ -0,0 +1,37 @@
+  molecular_model:
+    voxel_size: 0.1         # [nm]
```
geoffwoollard, line 2: are these string paths ever read by the TEM simulator? Or do we have freedom to change their names?
e.g. `voxel_size` to `pixel_size_ang`. If these are the "expected fields" to the TEM simulator, then I'm thinking we can have another layer above where standard params that are shared between different simulation routines are converted into "tem simulator" specific ones (units, etc).

temp_workspace/input/path_config.yaml
```diff
@@ -0,0 +1,5 @@
+  pdb_dir: '../temp_workspace/input/'
```
geoffwoollard, line 1: I thought we were going to get rid of a working directory... if it's something that only exists for the length of the simulation, perhaps it could be created and deleted with the `os` module

thisTyler: This particular working directory exists only for demonstration purposes - this path would be changed out by the user. temp_workspace/ should not be in the PR - I'll look into addressing this after work today.

temp_workspace/input/path_config.yaml
```diff
@@ -0,0 +1,5 @@
...
+  micrograph_keyword: '_randomrot'
+  pdb_keyword: '4v6x'
```
geoffwoollard, line 3: ok, so this would be an example .yaml file, and usually the user would have different hard coded values instead of `_randomrot` and `4v6x``

temp_workspace/README.md
```diff
@@ -0,0 +1,5 @@
+
+
+Temp workspace to test wrapper and keep input/output files.
```
geoffwoollard, line 3: I suggest you just delete this file

temp_workspace/input/path_config.yaml
thisTyler: Correct!

geoffwoollard (approved):

temp_workspace/input/path_config.yaml
ninamiolane: Maybe:
- having the pdb data in a `data/` or `pdb_structures/` directory inside `simSPI/simSPI`
- having the config files at the root of the `simSPI/simSPI` repo
would be more intuitive?

We do not need to tell the users how to organize their workspace.

ninamiolane (approved):
Looking good, thanks for doing the effort of splitting the work into block! 🙌  See my comment about the temp_workspace.

temp_workspace/input/path_config.yaml
ninamiolane: And the user can call the simulator by providing a path `output_dir` to tell it where to store the output of the simulation?

thisTyler: Good ideas, thank you!

I've edited the folder structure - temp_workspace is gone (as I mentioned earlier it's best not left in the main branch anyhow) and the config files now sit in the "root" simSPI/ directory. The PDBs are stored in simSPI/data as you suggested. 

For now, I think simSPI/output/ makes sense as a default output directory and the user can change it in the config file if desired.

temp_workspace/README.md
thisTyler: Agreed, done!

PR #18, opened 2021-11-23
Tem wrapper clean

jedyeo:


geoffwoollard:
You can run flake and black locally. Also remember to set up the hook so that it will format things for you. You just have to git add and git commit it again. Let me know if you are hitting snags on getting the style working.

simSPI/cryoemio.py
```diff
@@ -0,0 +1,392 @@
+import numpy as np
+import h5py
+import os
+import mrcfile
+import CifFile
+from CifFile import ReadCif
+
+
+## cryoSPARC related ##
```
ninamiolane, line 9: No leftover comments, put that in the docstring. --> please add docstrings (see comment on other PR)

simSPI/cryoemio.py
```diff
@@ -0,0 +1,392 @@
...
+    content = np.array(np.load(csfile).tolist())
+    dtype = np.load(csfile).dtype
+    header = []
+    for key in dtype.fields.keys():
+        header.append(key)
+    header = np.array(header)
+    return header, content
+
+
+### WIP !!!
```
ninamiolane, line 21: Ok, but there should be no WIP when the PR is ready for review

ninamiolane:
Sorry I started reviewing but I think this was not ready for review! This is an amazing body of code here, well done! 🎉 

One warning though: it seems that you are going very fast on the coding, but at the cost of having code that is not very clean (not documented and not tested 😭 )... :/ 

We will absolutely need **unit** tests for all of the functions, (i.e. one test per function). I would recommend to write the tests at the same time with the writing of the core code. 

You are probably testing your functions on simple examples when you code, e.g. just to ensure they run. These tests are the unit-tests and we need them.

Clean code is a difference between a code that you can use, and a code that everybody can use easily ! 

But I've looked at this PR too fast anyway, sorry again, I'll take another look when it's ready and you actually ask for my review 👋  Keep on the good work!! 🤩

fredericpoitevin:
Should we close this PR without merging @jedyeo @ninamiolane ?

jedyeo:
> Should we close this PR without merging @jedyeo @ninamiolane ?

yes

PR #19, opened 2021-11-23
Feature: adding helper functions from SLAC

jedyeo:
@calhep @thisTyler @arjunsingh3600 

This file contains some functions from the slac codebase that helps with the processing of particles. Tests won't cover this code yet but once we get the wrapper merged in (still cleaning that up) then we should be good.


simSPI/helpers.py
```diff
@@ -0,0 +1,218 @@
...
+import h5py
+import mdtraj as md
+import mrcfile
+import numpy as np
+from scipy.spatial.distance import pdist
+from scipy.stats import special_ortho_group
+
+
+def mrc2data(mrc_file=None):
+    """mrc2data."""
```
ninamiolane, line 14: We need a legitimate docstring, see our contributing guidelines:
https://github.com/compspi/compspi/blob/master/docs/contributing.rst#writing-docstrings

Can you add well-formatted docstrings to all of these functions? Thanks!

simSPI/helpers.py
```diff
@@ -0,0 +1,218 @@
...
+    if mrc_file is not None:
+        with mrcfile.open(mrc_file, "r", permissive=True) as mrc:
+            micrograph = mrc.data
+        if micrograph is not None:
+            if len(micrograph.shape) == 2:
+                micrograph = micrograph[np.newaxis, ...]
+        else:
+            print("Warning! Data in {} is None...".format(mrc_file))
```
ninamiolane, line 22: use the `logging` python package

simSPI/helpers.py
```diff
@@ -0,0 +1,218 @@
...
+
+import h5py
+import mdtraj as md
+import mrcfile
+import numpy as np
+from scipy.spatial.distance import pdist
+from scipy.stats import special_ortho_group
+
+
+def mrc2data(mrc_file=None):
```
ninamiolane, line 13: Why do we need this function to return None when nothing is given as input?

ninamiolane (changes requested):
Thank you for this PR! Great work!

High-level comments before I take another look:
- We absolutely need good docstrings: these are primordial for other users (including code reviewers) to understand what you are doing -- without spending 1 hour on it. If a user needs to spend more than 20 minutes to understand 10 lines of code, they will leave and go to another package.
- We also really need unit-tests. @jedyeo you said that you would put the tests once the wrapper is merged in, but we do not need the wrapper for the **unit** tests associated to this PR. Unit tests test one simple function at a time, they do **not** test the whole pipeline. We need test functions such as: `test_mrc2data, test_data_and_dic_2_hdf5`, etc, i.e. we need one test function per function implemented. This is why they are called "unit".
- It is not good practice to put many functions in a file called "helper.py", because "helper.py" does not say much about what is inside. some companies forbid the use of filenames such as `utils.py` or `helpers.py` just because people put everything in them. Please divide the code into `io.py` (for changing formats and write/load data) and `math.py` (for operations on rotations) files at least.

Could you address these? Then I can have a look at the code itself. Thank you!

simSPI/helpers.py
jedyeo: good question, i'll remove this (I made it return None to pass the lint, but let's just remove this structure and assume the user passes in correct input).

simSPI/crd.py
```diff
@@ -0,0 +1,135 @@
...
+    crd_file="crd.txt",
+):
+    """Write particle data to .crd file.
+
+    Parameters
+    ----------
+    numpart : int
+        Number of particles.
+    xrange : ndarray
+        Valid horizontal range for display.
```
geoffwoollard, line 22: units of pixels?

simSPI/crd.py
```diff
@@ -0,0 +1,135 @@
...
+    rotlist = []
+    for x in range(0, numpart + 1):
+        x = special_ortho_group.rvs(3)
+        y = rotation_matrix_to_euler_angles(x)
+        rotlist.append(y)
+    return rotlist
+
+
+def rotation_matrix_to_euler_angles(R):
+    """Compute Euler angles given a rotation matrix.
```
geoffwoollard, line 87: do you know what representation is assumed? ZYZ, ZXZ, etc.

https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    fov_Lx, fov_Ly, boxsize = get_fov(
+        optics_params,
+        detector_params,
+        pdb_file=pdb_file,
+        dmax=dmax,
+        pad=pad,
+    )
+
+    fov_Nx = np.floor(fov_Lx / boxsize)
```
geoffwoollard, line 35: NIT: use lowercase vs capital N and L

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+    optic_params : list
+        List of sim parameters pertaining to microscope settings.
+    detector_params : dict
+        List of sim parameters pertaining to detector settings.
+    pdb_file : str
+        Relative path to write .pdb file to.
+    dmax : int
+        Maximum dimension of molecule.
+    pad : int
+        Amount of padding.
```
geoffwoollard, line 25: mention units (pixels, angstrom, nm, etc)

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+    pdb_file : str
+        Relative path to .pdb file output.
+    dmax : int
+        Maximum dimension of molecule
+    pad : int
+        Amount of padding.
+    """
+    detector_Nx = detector_params[0]
+    detector_Ny = detector_params[1]
+    detector_pixel_size = detector_params[2] * 1e3
```
geoffwoollard, line 67: use units in the name here

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+def get_xyz_from_pdb(filename=None):
+    """Get particle coordinates from .pdb file.
+
+    Parameters
+    ----------
+    filename : str
+        Relative path to file containing topological information of particle
+    """
+    traj = md.load(filename)
+    atom_indices = traj.topology.select("name CA or name P")
```
geoffwoollard, line 106: can the user select all atoms, or are just the alpha carbons selected. Some tests that use the same atoms type (not realistic with real proteins) would be good to ensure what atoms are actually giving turned in to an image.

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+    n_ovl : int
+        Overlap.
+    """
+    if n_ovl == 0:
+        data_stack = blockshaped(data, n_boxsize, n_boxsize)
+    else:
+        n_split = math.floor((data.shape[0] - 2 * n_ovl) / (n_boxsize))
+        n_dilat = n_boxsize + 2 * n_ovl
+        data_stack = np.zeros((n_split * n_split, n_dilat, n_dilat))
+        print("Array dimensions: ", data_stack.shape)
```
geoffwoollard, line 163: log vs print

simSPI/helpers.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+    for k, v in dic.items():
+        if isinstance(v, (np.ndarray, np.int64, np.float64, int, float, str, bytes)):
+            h5file[path + k] = v
+        elif isinstance(v, type(None)):
+            h5file[path + k] = str("None")
+        elif isinstance(v, dict):
+            recursively_save_dict_contents_to_group(h5file, path + k + "/", v)
+        else:
+            raise ValueError("Cannot save %s type" % type(v))
```
geoffwoollard, line 72: I like this type of defensive coding, so we know when and why things fail. I had to move `assert False` to tests because no asserts were allowed in non-tests. I didn't realize that `raise valueError` was allowed (hopefully it really is).

simSPI/helpers.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+    rotlist = []
+    for x in range(0, numpart + 1):
+        x = special_ortho_group.rvs(3)
+        y = rotation_matrix_to_euler_angles(x)
+        rotlist.append(y)
+    return rotlist
+
+
+def rotation_matrix_to_euler_angles(R):
```
geoffwoollard, line 259: `rotation_matrix_to_euler_angles` is both here and in `crd.py`. Remove all duplicate code. Just import it from the same place instead of writing it again.

geoffwoollard (changes requested):
You need to have tests for basically every function. That way the automatic tests tell you where exactly something is failing. And you can also tests multiple expected inputs (big, small, etc).

Have a look at the tests for some of the code base. Then get writing and his that 90% and get those green checks!

simSPI/crd.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Parameters
+    ----------
+    R : ndarray
+        Rotation matrix.
+    """
+    if not is_rotation_matrix(R):
+        raise ValueError()
+
+    sy = math.sqrt(R[0, 0] * R[0, 0] + R[1, 0] * R[1, 0])
```
ninamiolane, line 97: No uppercase letter for variables. According to Python style guidelines (Pep8) uppercase letters are reserved to Objects of Classes. Use: rotation_mat instead.

simSPI/crd.py
```diff
@@ -0,0 +1,135 @@
...
+        x = math.atan2(-R[1, 2], R[1, 1])
+        y = math.atan2(-R[2, 0], sy)
+        z = 0
+    x = rad2deg(x)
+    y = rad2deg(y)
+    z = rad2deg(z)
+    return np.array([x, y, z])
+
+
+def rad2deg(x):
```
fredericpoitevin, line 113: I believe this is already available in numpy: https://numpy.org/doc/stable/reference/generated/numpy.rad2deg.html

Good news! Less unit tests to implement ;)

simSPI/crd.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Parameters
+    ----------
+    x : float
+        Degree value to convert to degrees.
+    """
+    return (x * 180) / np.pi
+
+
+def is_rotation_matrix(matrix):
```
fredericpoitevin, line 124: Might be simpler to just check that the determinant of the matrix is 1?
https://numpy.org/doc/stable/reference/generated/numpy.linalg.det.html

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+    detector_Lx = detector_Nx * detector_pixel_size
+    detector_Ly = detector_Ny * detector_pixel_size
+    fov_Lx = detector_Lx / magnification
+    fov_Ly = detector_Ly / magnification
+
+    if dmax is None:
+        if pdb_file is not None:
+            dmax = get_dmax(pdb_file)
+        else:
+            dmax = 100
```
fredericpoitevin, line 79: this could be simplified by setting dmax to 100 instead of None by default and overriding it if a pdb file is given.

simSPI/fov.py
```diff
@@ -0,0 +1,197 @@
...
+
+def get_xyz_from_pdb(filename=None):
+    """Get particle coordinates from .pdb file.
+
+    Parameters
+    ----------
+    filename : str
+        Relative path to file containing topological information of particle
+    """
+    traj = md.load(filename)
```
fredericpoitevin, line 105: I was very much into MDtraj when I wrote the material you used to make this function, but gemmi might be a better choice, since MDtraj is no longer maintained I think.

fredericpoitevin (commented):
I have added a few comments, but please first address Nina's concern about tests. Great work already everyone!!!

simSPI/crd.py
geoffwoollard: That's necessary, but not sufficient. we also need R=R.T. For instance [[a,a-1],[a+1,a]] has a det of 1 = a^2 - (a^2 -1), but is not a valid rotation matrix because it's not symmetric.

Now that I think of it, you should also check the determinant is one, not just that R=R.T, which is what the code does below.

```
is_symmetric = np.allclose(matrix,matrix.T)
is_det_one = np.isclose(np.linalg.det(matrix),1)
return is_symmetric and is_det_one
```

simSPI/io.py
```diff
@@ -0,0 +1,297 @@
...
+    beam_params=[300, 1.3, 100, 0],
+    dose=None,
+    optics_params=[81000, 2.7, 2.7, 50, 3.5, 0.1, 1.0, 0, 0],
+    defocus=None,
+    optics_defocout=None,
+    detector_params=[5760, 4092, 5, 32, "yes", 0.5, 0.0, 0.0, 1.0, 0, 0],
+    noise=None,
+    log_file="simulator.log",
+    seed=-1234,
+):
```
geoffwoollard, line 61: would it make more sense to have something here that read in a yaml specific file?

right now I have to look for a comment, then find the line of code that does that, and check the index of the list is correct. 

also the default values are in the comments. I'm scrolling up and down a lot to see if there is reasonable values.

I would re-write this to convert a yaml/config file to something that can be used conveniently to write the .inp file for the simulator

if the tem simulator takes the voxel size in um, but usually the pixel size is in angstroms, this could convert the units. and similarly for other parameters

simSPI/crd.py
jedyeo: Sorry, worded this really poorly. It's an array containing all x-coordinates of particle centers to write to the coordinate file.

jedyeo: changes reflected in the next push (coming soon)

simSPI/crd.py
calhep: Hi @geoffwoollard, just wanted to check whether the orthogonality condition for rotation matrices should be R • R.T = I instead of R = R.T.

geoffwoollard: Actually, I was wrong. a 3D rotation matrix is not in general symmetric!!!

Sorry XD

I'm not sure what all the conditions are called exactly, but we should definitely have
https://en.wikipedia.org/wiki/3D_rotation_group#Orthogonal_and_rotation_matrices

R dot Rt = Identity
(det R)**2 = 1

calhep: No worries! Thanks for clarifying.

simSPI/crd.py
```diff
@@ -0,0 +1,129 @@
...
+    """Determine whether a given matrix is a valid rotation matrix.
+
+    Parameters
+    ----------
+    matrix : ndarray
+        Matrix to check if it is a valid rotation matrix.
+    """
+    is_orthogonal = np.allclose(np.dot(matrix, matrix.T), np.identity(3))
+    is_det_one = np.isclose(np.linalg.det(matrix), 1)
+    return is_orthogonal and is_det_one
```
fredericpoitevin, line 129: 👍

simSPI/cryoemio.py
```diff
@@ -0,0 +1,314 @@
+"""Helper functions for tem.py processing of input and output files."""
```
fredericpoitevin, line 1: On a second read, this should probably be in ioSPI: what do you think?

simSPI/fov.py
fredericpoitevin: sorry did not notice this comment last time @geoffwoollard . I should not have hard-coded that in my notebooks in the past - this was only meant to deal with ribosomes in a reasonable amount of time. Ideally, you'd like the selection to be defined by the user - and by default all atoms should be selected.

tests/test_crd.py
```diff
@@ -0,0 +1,81 @@
+"""Unit tests for crd.py."""
```
fredericpoitevin, line 1: Nice guys!

tests/test_crd.py
```diff
@@ -0,0 +1,81 @@
...
+def test_write_crd_file():
+    """Test whether function creates and writes to file on given path."""
+    temp_dir = tempfile.TemporaryDirectory()
+    temp_file = temp_dir.name + "/crd.txt"
+
+    assert not os.path.isfile(temp_file)
+    crd.write_crd_file(numpart=20, crd_file=temp_file)
+    assert os.path.isfile(temp_file)
+    crd.write_crd_file(numpart=20, crd_file=temp_file)
+    assert os.path.isfile(temp_file)
```
fredericpoitevin, line 20: Why is this done twice?

tests/test_crd.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_get_rotlist():
+    """Test whether randomly generated rotlist is of correct length."""
+    numpart = 10
+    assert len(crd.get_rotlist(numpart)) == len(range(0, numpart + 1))
+
+
+def test_rotation_matrix_to_euler_angles():
```
fredericpoitevin, line 29: this looks good to me

fredericpoitevin (commented):
Thanks for adding tests, they look great!

Looking into the build summary, I see a lot of `mdtraj` related warnings. This library is not maintained anymore unfortunately. Nothing urgent, but in time we will need to replace it with `gemmi`.

My suggestion to get this to pass is to move `cryoemio.py` to ioSPI - it will boost your coverage statistics:
```----------- coverage: platform linux, python 3.9.5-final-0 -----------
Name                   Stmts   Miss  Cover   Missing
----------------------------------------------------
simSPI/__init__.py         1      0   100%
simSPI/crd.py             49      0   100%
simSPI/cryoemio.py        96     96     0%   2-300
simSPI/fov.py             79      0   100%
simSPI/multislice.py      21      0   100%
simSPI/tem.py             46      1    98%   199
simSPI/transfer.py        50      3    94%   94, 101, 171
----------------------------------------------------```

File: simSPI/fov.py (hunk 17 above)
Comment (jedyeo) [reply simSPI/fov.py:134]: if you're okay with it, let's leave this for a future PR?

File: simSPI/cryoemio.py (hunk 20 above)
Comment (jedyeo) [reply simSPI/cryoemio.py:1]: my understanding is that ioSPI is reserved for processes related to data flowing between user and OSF cloud - maybe might be best to get more input but I'd prefer to leave it in simSPI for the time being unless others think different.

Comment (jedyeo) [reply simSPI/cryoemio.py:1]: (also tests are coming soon for cryoemio.py)

File: simSPI/crd.py (hunk 5 above)
Comment (jedyeo) [reply simSPI/crd.py:90]: XYZ

File: tests/test_crd.py (hunk 22 above)
Comment (jedyeo) [reply tests/test_crd.py:20]: tests writing to a file that already exists, should we add a comment to clarify?

File: simSPI/cryoemio.py (hunk 20 above)
Comment (fredericpoitevin) [reply simSPI/cryoemio.py:1]: The I/O in ioSPI refers to https://en.wikipedia.org/wiki/Input/output
and the purpose of that repository is to have all necessary functions and methods to read and write any type of file that contains data, results, etc.
In that sense, I think the functions here belong in ioSPI.

File: simSPI/fov.py (hunk 17 above)
Comment (fredericpoitevin) [reply simSPI/fov.py:134]: Sounds good. I created the corresponding issue #22

**Review by fredericpoitevin — CHANGES_REQUESTED** [review]
Let's move cryoemio.py to ioSPI. See my answer in comments - and let's not forget this will likely address the coverage problem.

**Review by fredericpoitevin — COMMENTED** [review]
Great work everyone @calhep @jedyeo @thisTyler @arjunsingh3600 ! I think this PR is getting really close to being merged!

Pending Nina's approval, I think you just need to address the test that did not pass (it has to do with `multislice.py` and only happened on one python version - maybe have a quick look?) and it can be merged.

I have one question about the `cryoemio.py` functions that I have asked you to move to ioSPI (thanks I see that you are working on the PR there, let us know when it's ready to review): what was the initial motivation for having it here? I don't see any of these functions used in the code. Thanks for enlighting me :D

**Comment by thisTyler** [comment]
> @fredericpoitevin: what was the initial motivation for having it here? I don't see any of these functions used in the code.
These functions were originally sourced from simutils.py in the cryoEM-notebooks repository and were used in your notebook interfacing the TEM simulator. This was the original impetus, as we intend to use a similar workflow for the wrapped simulator. Hope this helps! :)

File: simSPI/fov.py [hunk 24]
```diff
@@ -0,0 +1,249 @@
...
+    micrograph, optics_params, detector_params, pdb_file=None, dmax=30, pad=5.0
+):
+    """Extract particles from given micrograph.
+
+    Parameters
+    ----------
+    micrograph : ndarray
+        Particle micrograph to extract from.
+    optics_params : list
+        List of sim parameters pertaining to microscope settings.
```
ninamiolane, line 154: Replace by `Simulations parameters related to microscope settings`
(avoid using the abbreviation sim, and we don't need to specify that it is a list, it is said above)

To be changed here and elsewhere.

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Returns
+    -------
+    data_stack : ndarray
+        Array containing sliced particle data.
+    """
+    log = logging.getLogger()
+
+    if n_ovl == 0:
+        data_stack = blockshaped(data, n_boxsize, n_boxsize)
```
ninamiolane, line 204: You can directly do:
`return blockshaped(data, n_boxsize, n_boxsize)`
which spares you the `else` after

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+    else:
+        n_split = math.floor((data.shape[0] - 2 * n_ovl) / n_boxsize)
+        n_dilat = n_boxsize + 2 * n_ovl
+        data_stack = np.zeros((n_split * n_split, n_dilat, n_dilat))
+        log.info("Array dimensions: {}".format(data_stack.shape))
```
ninamiolane, line 209: Use f"Array dimensions: {data_stack.shape}"

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+    """Convert a 2D numpy array into a 3D numpy array.
+
+    Parameters
+    ----------
+    data : ndarray
+        2D array to stack.
+    n_boxsize : int
+        Boxsize.
+    n_ovl : int
+        Overlap.
```
ninamiolane, line 194: What does overlap means for this function? Could you explain a bit in the docstring?

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+                rows = np.arange(istart, istop)
+                columns = np.arange(jstart, jstop)
+                data_tmp = data[np.ix_(rows, columns)]
+                data_stack[i_stack, ...] = data_tmp[np.newaxis, ...]
+                i_stack += 1
+
+    return data_stack
+
+
+def blockshaped(arr, nrows, ncols):
```
ninamiolane, line 226: For consistency of variable names: `arr` -> `data`?

ninamiolane: How is this function different from just doing `arr.reshape(-1, nrows, ncols)`? What is the expected input shape of the arr?

tests/test_crd.py
```diff
@@ -0,0 +1,81 @@
...
+    crd.write_crd_file(numpart=20, crd_file=temp_file)
+    assert os.path.isfile(temp_file)
+    crd.write_crd_file(numpart=20, crd_file=temp_file)
+    assert os.path.isfile(temp_file)
+
+
+def test_get_rotlist():
+    """Test whether randomly generated rotlist is of correct length."""
+    numpart = 10
+    assert len(crd.get_rotlist(numpart)) == len(range(0, numpart + 1))
```
ninamiolane, line 26: Putting `== numpart` directly would be more readable?

also: numpart -> n_particles

jedyeo:
>what was the initial motivation for having it here? I don't see any of these functions used in the code. Thanks for enlighting me :D

To expand on Tyler's answer, the wrapper will be utilizing these functions. I think the current code in whatever branch our wrapper implementation lies in does not reference it as cryoemio, rather as io. We changed it to cryoemio because io is already a default library in python :)

tests/test_crd.py
ninamiolane: Very nice example of testing!

ninamiolane: (we might have to delete it though, since we might be able to use the scipy.spatial function to deal with rotation matrices and euler angles: we should not reimplement what already exists elsewhere)

tests/test_crd.py
```diff
@@ -0,0 +1,81 @@
...
+    theta_s = np.arctan2(-s_rot[1, 2], s_rot[1, 1])
+    psi_s = np.arctan2(-s_rot[2, 0], 0)
+    phi_s = 0
+    s_angles = np.array([np.rad2deg(theta_s), np.rad2deg(psi_s), np.rad2deg(phi_s)])
+
+    assert np.allclose(ns_angles, crd.rotation_matrix_to_euler_angles(ns_rot))
+    assert np.allclose(s_angles, crd.rotation_matrix_to_euler_angles(s_rot))
+
+
+def test_rotation_matrix_to_euler_angles_exception():
```
ninamiolane, line 56: Very nice testing

tests/test_fov.py
```diff
@@ -0,0 +1,127 @@
...
+    # optics params
+    magnification = 81000
+
+    x, y, n = fov.define_grid_in_fov(
+        [magnification], [detector_nx, detector_ny, detector_pixel_size], pdb_path
+    )
+
+    assert x.size == 16
+    assert y.size == 11
+    assert n == 176
```
ninamiolane, line 38: Nice

ninamiolane (changes requested):
Very nice, excellent work and excellent testing too! 🎉 

I made several comments to be sure we start with very clean code, but the code was already very nice. Just some functions that we might not need (it seems that they already exist in other open source packages such as scipy.spatial) and some inconsistency in variable names (being consistent in variable naming is important, as users will get very frustrated if we use `numpart` in some functions but `n_particles` in others, since they have to check each time which one it is to use our function.)

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
+"""Helper functions for tem.py to define grid FOV for displaying particles."""
```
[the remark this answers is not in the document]
fredericpoitevin: (in case nobody knows the answer, I think I intended it to mean "Field Of View" - thank you guys for putting up with my messy notebooks and making such a nice piece of code!)

tests/test_crd.py
geoffwoollard: you could write it in code like 

```
file_already_exists = temp_file
crd.write_crd_file(numpart=20, crd_file=file_already_exists)
```

also, the `assert os.path.isfile(temp_file)` in line 20 would already be written from line 17, and not necessarily 19.
you would want to make a change in the actual file, use the same file name, and check that the new change inside the file that got written to disk is there.

simSPI/crd.py
```diff
@@ -0,0 +1,129 @@
...
+        Relative path to output .crd file.
+    """
+    log = logging.getLogger()
+
+    if os.path.exists(crd_file):
+        log.info(crd_file + " already exists.")
+    else:
+        rotlist = get_rotlist(numpart)
+        with open(crd_file, "w") as crd:
+            crd.write("# File created by TEM-simulator, version 1.3.\n")
```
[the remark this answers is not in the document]
geoffwoollard: Good idea. Perhaps running the tem simulator with --version or something, and capturing the stout?

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+
+import mdtraj as md
+import numpy as np
+from scipy.spatial.distance import pdist
+
+
+def define_grid_in_fov(
+    optics_params, detector_params, pdb_file=None, dmax=None, pad=1.0
+):
+    """Define particle grid for picking particles from micrograph.
```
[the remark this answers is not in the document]
geoffwoollard: I think particles and maybe "micrographs" will eventually be referred to in many files in simSPI and other compSPI repos. Do we want to be explaining particles and micrographs every time they are referred to, or in one main high level place?

tests/test_crd.py
geoffwoollard: You can also just use more explicit variable names instead of comments like `# generate sample non-singular (ns) and singular (s) rotation matrices`, e.g. 
```
nonsingular_rot = np.array([[1, 0, 0], [0, c, -s], [0, s, c]])
singular_rot = np.array([[0, 0, 1], [0, 1, 0], [-1, 0, 0]])
```

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+    fov_ly = detector_ly / magnification
+
+    if pdb_file is not None:
+        dmax = get_dmax(pdb_file)
+
+    boxsize = dmax + 2 * pad
+    return fov_lx, fov_ly, boxsize
+
+
+def get_dmax(filename):
```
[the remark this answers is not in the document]
geoffwoollard: `pdf_file` or `pdb_file`?

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+        Max dimension of given particle in .pdb source file.
+    """
+    xyz = get_xyz_from_pdb(filename)
+    distance = pdist(xyz[0, ...])
+    dmax = np.amax(distance)
+
+    return dmax
+
+
+def get_xyz_from_pdb(filename=None, atom_selection="name CA or name P"):
```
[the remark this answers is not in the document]
geoffwoollard: For now I would not do `name CA` as a default as that will only select the one atom (the alpha carbon) per amino acid. I would make it more general, perhaps `name P` which probably means only protein (and not nucleic acid).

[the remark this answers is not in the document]
geoffwoollard: See https://github.com/compSPI/simSPI/issues/22#issuecomment-988115921

simSPI/fov.py
geoffwoollard: > How is this function different from just doing arr.reshape(-1, nrows, ncols)? What is the expected input shape of the arr?

the swapaxes moves data between different axes.

tests/test_fov.py
```diff
@@ -0,0 +1,127 @@
...
+)
+
+pdb_dir = tempfile.TemporaryDirectory()
+pdb_path = pdb_dir.name + "/fov_test.pdb"
+
+with open(pdb_path, "w") as f:
+    f.write(pdb_test_string)
+
+
+def test_define_grid_in_fov():
```
geoffwoollard, line 22: if you did any work on paper figuring out these numbers, and know the equations, please put them in the doc string (vs just ran things and looked at the output). if you don't know exactly where they come from, that's OK.

geoffwoollard (changes requested):
Good job. Some of the tests were very readable.

simSPI/crd.py
```diff
@@ -0,0 +1,129 @@
+"""Helper functions for tem.py to write particle data to .crd file."""
```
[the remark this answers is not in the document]
jedyeo: let me know if you like the changed docstring

simSPI/crd.py
```diff
@@ -0,0 +1,129 @@
...
+from scipy.stats import special_ortho_group
+
+
+def write_crd_file(
+    numpart,
+    xrange=np.arange(-100, 110, 10),
+    yrange=np.arange(-100, 110, 10),
+    crd_file="crd.txt",
+):
+    """Write particle data to .crd file.
```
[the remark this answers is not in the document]
jedyeo: see new changes and let me know if it is acceptable :)

simSPI/fov.py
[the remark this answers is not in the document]
fredericpoitevin: `name CA or name P` means that C-alpha and Phosphate atoms will be selected. The former can be seen as a way to coarse-grain proteins, the latter nucleic acids.

simSPI/fov.py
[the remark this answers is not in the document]
fredericpoitevin: See newly created #31

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+    """Get particle coordinates from .pdb file.
+
+    Parameters
+    ----------
+    filename : str
+        Relative path to file containing topological information of particle
+    atom_selection: str
```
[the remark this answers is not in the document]
fredericpoitevin: `name CA or name P` means that C-alpha and Phosphate atoms will be selected. The former can be seen as a way to coarse-grain proteins, the latter nucleic acids.

Maybe enough to reference this link for now? https://mdtraj.org/1.9.4/examples/atom-selection.html#Atom-Selection-Language

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+        Atoms to be selected by MDTraj for returning their cartesian coordinates
```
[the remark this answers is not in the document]
fredericpoitevin: Maybe adding this would be enough?

```
MDtraj is a python library that allows to read, write and analyze MD trajectories with only a few lines of Python code.
https://mdtraj.org/1.9.4/index.html
```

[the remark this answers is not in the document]
fredericpoitevin: See #22 : `gemmi` would be a better library for the task at hand here, namely reading PDB files, and it is still maintained while `MDtraj` is not anymore.

simSPI/crd.py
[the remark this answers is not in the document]
jedyeo: Given the imminent deadline, let's leave it for a future PR.

simSPI/fov.py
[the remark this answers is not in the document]
jedyeo: shall we leave as is or change to `name P`?

simSPI/fov.py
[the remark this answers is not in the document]
jedyeo: added in docstring for clarification

simSPI/fov.py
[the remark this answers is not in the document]
jedyeo: i added in a comment with a link to selection syntax, would that be sufficient?

simSPI/fov.py
```diff
@@ -0,0 +1,249 @@
...
+    -------
+    particles : ndarray
+        Picked and sliced particle data from micrograph.
+    """
+    fov_Lx, fov_Ly, boxsize = get_fov(
+        optics_params, detector_params, pdb_file=pdb_file, dmax=dmax, pad=pad
+    )
+    fov_Nx = np.floor(fov_Lx / boxsize)
+    fov_Ny = np.floor(fov_Ly / boxsize)
+    pixel_size = (fov_Lx / micrograph.shape[1] + fov_Ly / micrograph.shape[0]) / 2.0
```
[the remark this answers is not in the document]
jedyeo: yes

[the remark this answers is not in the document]
jedyeo: yes

simSPI/fov.py
jedyeo: to be honest i am not sure - @fredericpoitevin maybe you can elaborate more? that else statement is never executed because `n_ovl` is always 0, so maybe we can just remove that part?

jedyeo:
@fredericpoitevin i changed the convention to ZYZ!

simSPI/crd.py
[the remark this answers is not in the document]
geoffwoollard: https://github.com/compSPI/simSPI/issues/33

geoffwoollard (approved):
let's push this PR through and see how things come together with the unit tests with the other PRs https://github.com/compSPI/ioSPI/pull/24 https://github.com/compSPI/simSPI/pull/32

If there are anymore issues that came up with future TODOs, I suggest you make issues of them so the ideas don't get lost track of

fredericpoitevin:
@jedyeo I think you changed to "zyz" not "ZYZ", right?

@geoffwoollard looks like tests are not passing...

arjunsingh3600:
@fredericpoitevin to address the failed tests, would it be better to revert the PR or to create a new PR addressing the tests?

jedyeo:
> @fredericpoitevin to address the failed tests, would it be better to revert the PR or to create a new PR addressing the tests?

create new pr is best move

fredericpoitevin:
@jedyeo and @arjunsingh3600 I'm not sure I agree. Would have been preferable to fix things in this PR prior merging. Now the problem will spread to other branches being created before it is solved.

arjunsingh3600:
@fredericpoitevin I've created a revert PR request [here](https://github.com/compSPI/simSPI/pull/39). This should remove the code from the main branch while I push the test fixes.

fredericpoitevin:
Thanks @arjunsingh3600 great initiative!
Right now I am going to merge this "new" revert PR #39 
Then you can open a new PR that would essentially replicate what this "old" buggy PR #19 was; and we can work on addressing the problems from there.

Issue #14, opened 2021-11-13
simulator inputs and outputs: .config, meta data, ground truth for weak-phase-simulator and TEM-simulator


geoffwoollard:
# Design desiderata

A simulated dataset should be fully reproducible
* input: simulation method, all input params needed for the simulation method, specimen, random seed
* output: particle stack, meta-data

The output meta-data (star file) should be unified to come extent across different simulation methods
* optics parameters (CTF parameters)
* pixel size

Different simulators take different input parameters
* amount of Gaussian white noise in the weak-phase-simulator
* detector effects (MTF) in the TEM simulator

# Proposed design

1. input .config specific to the simulator
  * e.g. weak phase: https://github.com/harshit-gupta-cor/simSPI/blob/wpsimulator/simSPI/configs/parameters_generate_dataset.cfg
  * e.g. TEM simulator (LINK FORTHCOMING FROM CAPSTONE STUDENTS)
 
2. shared common .config
* optics
  * CTF
    * defocus_1_ang
    * defocus_2_ang
    * defocus_angle_rad
    * etc.
  * electron_dose_e_ang2
* simulator method
  * e.g. "weak_phase" or "tem_simulator"
* simulator method specific .config file (see 1.)
* pixel_size_ang

3. Meta data returned in star file with ground truth values from the simulation method
* Categories that are specific to the simulation method (e.g. ground truth of detector or ice thickness effects in the TEM simulator) are prefixed with the simulation method. Note that the input params in 1. may get transformed and produce output values that are only known after run time.
* Categories shared between all simulation methods have no method specific prefix

geoffwoollard:
# Common input/output categories

* pixel_size_ang # pixel size in angstroms
* voltage_kv # microscope beam voltage, e.g. 300 keV
* electron_dose_e_ang2 # electron dose in electrons / Angstrom^2
* spherical_aberration_mm # spherical aberration in milimeters
* defocus_1_ang # defocus 1 in angstroms
* defocus_2_ang # defocus 2 in angstroms
* astigmatism_angle_rad # defocus angle of astigmatism  in radians

# Weak phase approximation specific categories
  * amplitude_contrast # a ratio to no units
[...]

PR #20, opened 2021-11-28
Added linear simulator and test functions

harshit-gupta-cor:
Added the linear simulator and pdated README.md on how to run the linear simulator. It comes with test functions for the individual modules.


README.md
```diff
@@ -11,11 +11,95 @@ Methods and tools for simulating SPI data.
...
 
+
+
+# Linear Simulator
+There are two types of simulator in this package:
+Linear and TEM.
+
+The linear simulator uses the linear model of the Cryo-EM physics to generate data.
+This involves fist rotating the biomolecule and obtaining its 2D tomographic projection.
+This 2D projection is then modulated with a CTF (Contrast Transfer Function) in fourier domain (convolution with corresponding PSF in image domain) and then shifted using scpecified shift values.
```
geoffwoollard, line 38: shifted in real or fourier?

environment.yml
```diff
@@ -4,19 +4,24 @@ channels:
...
   - pytest-cov
+  - pytorch=1.9.0
   - pip:
     - black==20.8b0
     - geomstats
     - git+https://github.com/compSPI/ioSPI.git
     - pre-commit
-    - raster_geometry
\ No newline at end of file
+    - pytorch3d==0.5.0
```
geoffwoollard, line 25: there seems to be some issue with `pytorch3d` in the linting/testing

simSPI/linear_simulator/config.py
```diff
@@ -0,0 +1,64 @@
...
+        self.volumedomain = config.get(s, "volumedomain").lower()
+        self.sidelen = config.getint(s, "sidelen")
+        self.angle_distribution = config.get(s, "angle_distribution").lower()
+        self.relion_invert_hand = config.getboolean(s, "relion_invert_hand")
+
+        s = "ctf"
+        self.ctf = config.getboolean(s, "ctf")
+        self.changectfs = config.getboolean(s, "changectfs")
+        self.ctf_size = config.getint(s, "ctf_size")
+        self.valueNyquist = config.getfloat(s, "valueNyquist")
```
geoffwoollard, line 47: NIT: no camel case. Use `value_nyquist`

simSPI/linear_simulator/config.py
```diff
@@ -0,0 +1,64 @@
...
+        self.bfactor = config.getfloat(s, "bfactor")
+        self.pixel_size = config.getfloat(s, "pixel_size")
+        self.kV = config.getfloat(s, "kV")
```
geoffwoollard, line 50: `kv`

simSPI/linear_simulator/config.py
```diff
@@ -0,0 +1,64 @@
...
+        self.cs = config.getfloat(s, "cs")
+        self.amplitude_contrast = config.getfloat(s, "amplitude_contrast")
+        self.min_defocus = config.getfloat(s, "min_defocus")
+        self.max_defocus = config.getfloat(s, "max_defocus")
```
geoffwoollard, line 54: I suggest using units in variable names. e.g. `min_defocus_ang` or `min_defocus_um`

simSPI/linear_simulator/configs/betagal_distribution.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # int: (default: 1)   real size of the pixel of the detector in Angstrom (A)
+    kV= 300
+            # float: (default: 300) voltage in kV
+    valueNyquist= 0.2
+            # float: (default: 0.2) value of the envelope at the nyquist frequency
+    bfactor=0
+            # float: (default: 0) value of the envelope B-factor
+
+    cs=2.7
+            # float: (default :2.7) spherical abberation in mm
```
geoffwoollard, line 46: cs_mm

simSPI/linear_simulator/configs/betagal_distribution.cfg
```diff
@@ -0,0 +1,71 @@
...
+    amplitude_contrast= 0.1
+            # float: (default :0.1)
+    min_defocus=1
+            # float: (default :1) min defocus in um for generating the ctf parameters values when starfile is not given
+    max_defocus=2
```
geoffwoollard, line 51: min_defocus_um, max_defocus_um

simSPI/linear_simulator/configs/betagal_distribution.cfg
```diff
@@ -0,0 +1,71 @@
...
+    astigmatism=False
+            # bool: (default: False) option to enable astigmatism #TODO: to be implemented
+
+[shift]
+    shift=True
+            # bool: (default: True) option to enable shifts
+    shift_distribution=triangular
+            # str: (default: triangular) name of the distribution from which the shifts are sampled
+    shift_variance=5
+            # float: variance of the shift distribution
```
geoffwoollard, line 62: shift_std_pix.

std is more natural than variance because its the same units as the distributions random variable. Also good to mention units. I assume its pixels here and not ang, etc.

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
+[general]
+    name = betagal_starfile
+            #str: name of the dataset folder
+    usegpu= False
+            #bool: (default: False) option to use GPU TODO: gpu option to be added
+    datasetSize= 100
```
geoffwoollard, line 6: num_particles

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            #int: size of the dataset
+    chunks=10
```
geoffwoollard, line 8: num_chunks or chunck_size

`chunks` seems like num_chunks, but the comments suggests chunck_size

README.md
```diff
@@ -11,11 +11,95 @@ Methods and tools for simulating SPI data.
...
+There are two main mode of generating the dataset:
+
+
+_Starfile-based_
+
+In this mode the parameters of the linear simulator for each projection are obtained from a specified star file whose path is given in the config file.
+
+_Distributional_
+
+In this mode the parameters of the linear simulator for each projection are obtained using specified distribution.
```
ynashed, line 78: randomly from uniform distributions (or something more specific)

README.md
```diff
@@ -11,11 +11,95 @@ Methods and tools for simulating SPI data.
...
+To choose this mode, leave the input star file path in config file blank.
+
+
+**Example Scripts**
+
+Example scripts are given in _example_scripts_ folder
+
+To run them move to base folder of the directory and run, for example,
+
+    ./example_scripts/generate_betagal_data_distributional.sh
```
ynashed, line 88: I don't see the necessity of these bash scripts if they only run a python module.

example_scripts/generate_betagal_data_distributional.sh
```diff
@@ -0,0 +1,3 @@
+ #script that runs dataset generator with imaging parameters coming from distribution specified in the .cfg file
+python -m simSPI.linear_simulator.main simSPI/linear_simulator/configs/betagal_distribution.cfg
```
ynashed, line 2: Consider removing these bash scripts.

example_scripts/generate_betagal_data_from_starfile.sh
```diff
@@ -0,0 +1,3 @@
+ #script that runs dataset generator with imaging parameters coming from a starfile
+ python -m simSPI.linear_simulator.main simSPI/linear_simulator/configs/betagal_starfile.cfg
+
```
ynashed, line 3: Consider removing these bash scripts.

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # bool: (default: True) option to enable ctf convolution
+    changectfs = True
+            # bool: (default: True) option to change ctf per image
+    ctf_size = 64
+            # int: (default: str) size of the ctf in pixels. Currently, in the code this is forecefully made equal to sidelen
+    pixel_size= 1
+            # int: (default: 1)   real size of the pixel of the detector in Angstrom (A)
+    kV= 300
+            # float: (default: 300) voltage in kV
+    valueNyquist= 0.2
```
geoffwoollard, line 40: value_nyquist

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+[ctf]
+    ctf = True
+            # bool: (default: True) option to enable ctf convolution
+    changectfs = True
+            # bool: (default: True) option to change ctf per image
+    ctf_size = 64
+            # int: (default: str) size of the ctf in pixels. Currently, in the code this is forecefully made equal to sidelen
+    pixel_size= 1
+            # int: (default: 1)   real size of the pixel of the detector in Angstrom (A)
+    kV= 300
```
geoffwoollard, line 38: kv

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # float: (default: 300) voltage in kV
+    valueNyquist= 0.2
+            # float: (default: 0.2) value of the envelope at the nyquist frequency
+    bfactor=0
+            # float: (default: 0) value of the envelope B-factor
+
+    cs=2.7
```
geoffwoollard, line 45: cs_mm

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # float: (default :2.7) spherical abberation in mm
+    amplitude_contrast= 0.1
+            # float: (default :0.1)
+    min_defocus=1
```
geoffwoollard, line 49: min_defocus_ang

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # float: (default :1) min defocus in um for generating the ctf parameters values when starfile is not given
+    max_defocus=2
```
geoffwoollard, line 51: max_defocus_um

simSPI/linear_simulator/configs/betagal_starfile.cfg
```diff
@@ -0,0 +1,71 @@
...
+            # float: (default :2) max defocus in um for generating the ctf parameters values when starfile is not given
+    astigmatism=False
+            # bool: (default: False) option to enable astigmatism #TODO: to be implemented
+
+[shift]
+    shift=True
+            # bool: (default: True) option to enable shifts
+    shift_distribution=triangular
+            # str: (default: triangular) name of the distribution from which the shifts are sampled
+    shift_variance=5
```
geoffwoollard, line 61: shift_std_pix or shift_std_ang

simSPI/linear_simulator/dataset_generator.py
```diff
@@ -0,0 +1,139 @@
...
+        self.print_statements()
+        self.params_generator = ParamsFactory.get_params_generator(
+            config
+        )  # instantiating params generator
+
+    def run(self):
+        """Generate a chunk of projection and save it."""
+        for iterations in range(
+            self.config.datasetSize // self.config.chunks
+        ):  # TODO: last few images will be left
```
geoffwoollard, line 42: you can use something like https://numpy.org/doc/stable/reference/generated/numpy.array_split.html to get pieces that are roughly the right size

or get indices and do [stard:end], check that end < len and do a final [start:]

simSPI/linear_simulator/dataset_generator.py
```diff
@@ -0,0 +1,139 @@
...
+            rot_params,
+            ctf_params,
+            shift_params,
+            iterations,
+            self.config,
+        )
+
+    def print_statements(self):
+        """Print statements about the data."""
+        print(f"The size of the dataset is {self.config.datasetSize}")
```
geoffwoollard, line 133: log instead of print

simSPI/linear_simulator/dataset_generator.py
```diff
@@ -0,0 +1,139 @@
...
+        for iterations in range(
+            self.config.datasetSize // self.config.chunks
+        ):  # TODO: last few images will be left
+            rot_params, ctf_params, shift_params = self.params_generator.get_params()
+            projections = self.simulator(rot_params, ctf_params, shift_params)
+            self.save_mrc(projections, iterations)
+            self.starfile_data(rot_params, ctf_params, shift_params, iterations)
+        self.save_starfile()
+        self.save_configfile()
+
```
geoffwoollard, line 49: I suggest writing .mrc files of a reasonable size, and breaking up into batches if they get too big. 

As written, won't this overwrite "Simulated.star"?

simSPI/linear_simulator/main.py
```diff
@@ -0,0 +1,41 @@
...
+    dataset_generator = dataset_gen(config)
+    dataset_generator.run()
+    return 0, "Dataset successfully generated."
+
+
+if __name__ == "__main__":
+    try:
+        retval, status_message = main()
+    except Exception as e:
+        print(traceback.format_exc(), file=sys.stderr)
```
geoffwoollard, line 36: log

simSPI/linear_simulator/main.py
```diff
@@ -0,0 +1,41 @@
...
+        retval = 1
+        status_message = "Error: Training failed."
+
+    print(status_message)
```
geoffwoollard, line 40: log

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+
+    """
+
+    def __init__(self, config, initial_volume=None):
+        super(simulator, self).__init__()
+
+        self.config = config
+        self.Projector = Projector(config)  # class for tomographic projector
+        self.init_volume()  # changes the volume inside the projector
+        self.CTF = CTF(config)  # class for ctf
```
geoffwoollard, line 31: I'm not exactly sure what naming case this falls into https://www.python.org/dev/peps/pep-0008/#naming-conventions

Are you sure it should be 
```
self.Projector = Projector(config)
...
self.CTF = CTF(config)
```
and not 
```
self.projector = Projector(config)
self.ctf = CTF(config)
```

geoffwoollard: Nevermind, I see how these are still Classes, with there own methods that need to be called.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        ctf_params: dict of type str to {tensor}
+            Dictionary of Contrast Transfer Function (CTF) parameters
+             for a projection chunk
+        shift_params: dict of type str to {tensor}
+            Dictionary of shift parameters for a projection chunk
+
+        Returns
+        -------
+        projection.real : tensor
+            Tensor ([chunks,1,sidelen,sidelen]) contains cryoEM measurement
```
geoffwoollard, line 51: is this the fastest way to do things vs [sidelen,sidelen,1,chunks]? What's the fast axis and slow axis?

Also, why the 4 dimensions, with the 1 in [chunks,1,sidelen,sidelen]?

simSPI/linear_simulator/modules.py
ynashed: @geoffwoollard I think you were correct earlier. it should be `self.projector` as `projector` is a class instance variable not a class name.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+            rot_vol.reshape(
+                batch_sz, self.config.sidelen, self.config.sidelen, self.config.sidelen
+            ),
+            dim=proj_axis,
+        )
+        projection = projection[
+            :, None, :, :
+        ]  # add a dummy channel (for consistency w/ img fmt)
+        # --> B,C,H,W
+        return projection
```
geoffwoollard, line 151: is there any masking? are there sharp edges from the cube? even if you can't see them with noise, we don't want them there since it gives information about the rotation that is very artificial. perhaps there could be a soft spherical masking of the initial input volume that goes all the way out to Nyquist.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+
+    def _get_ewavelength(self):
+        """Output the wavelength of the electron beam.
+
+        Returns
+        -------
+        wavelength: float
+            wavelength of the electron beam
+        """
+        wavelength = 12.2639 / np.sqrt(
```
geoffwoollard, line 188: Some more/different sig figs from https://github.com/asarnow/pyem/blob/master/pyem/ctf.py

12.2643247
0.978466

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+            self.config.kV * 1e3 + 0.97845 * self.config.kV ** 2
```
geoffwoollard, line 189: self.config.kV * (1e3 + 0.97845 * self.config.kV)

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        defocusV = ctf_params["defocusV"]
+        angleAstigmatism = ctf_params["defocusAngle"]
+        elliptical = (
+            defocusV * self.r2
+            + (defocusU - defocusV)
+            * self.r2
+            * torch.cos(self.angleFrequency - angleAstigmatism) ** 2
+        )
+        defocusContribution = (
+            np.pi * self.wavelength * 1e4 * elliptical * self.frequency ** 2
```
geoffwoollard, line 217: can reuse `self.frequency ** 2` in `self.frequency ** 4`

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        )
+        abberationContribution = (
+            -np.pi
+            / 2.0
+            * self.config.cs
+            * (self.wavelength ** 3)
+            * 1e7
+            * self.frequency ** 4
+            * self.r2 ** 2
+        )
```
geoffwoollard, line 227: math looks good

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        defocusU = ctf_params["defocusU"]
+        defocusV = ctf_params["defocusV"]
+        angleAstigmatism = ctf_params["defocusAngle"]
+        elliptical = (
+            defocusV * self.r2
+            + (defocusU - defocusV)
+            * self.r2
+            * torch.cos(self.angleFrequency - angleAstigmatism) ** 2
+        )
+        defocusContribution = (
```
geoffwoollard, line 216: math looks good. at least up to that factor of 1e4 in line 217.

if you had units in the variable names, it would make it easier to verify.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+            * self.config.cs
+            * (self.wavelength ** 3)
+            * 1e7
+            * self.frequency ** 4
+            * self.r2 ** 2
+        )
+
+        argument = abberationContribution + defocusContribution
+
+        hFourier = (1 - self.config.amplitude_contrast ** 2) ** 0.5 * torch.sin(
```
geoffwoollard, line 231: do you have the sign right? the CTF is negative close to zero, and the amplitude contrast makes it (more) negative at zero frequency?

geoffwoollard: https://www.youtube.com/watch?v=Y8wivQTJEHQ&t=1833s

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        -------
+        wavelength: float
+            wavelength of the electron beam
+        """
+        wavelength = 12.2639 / np.sqrt(
+            self.config.kV * 1e3 + 0.97845 * self.config.kV ** 2
+        )
+        return wavelength
+
+    def get_ctf(self, ctf_params):
```
geoffwoollard, line 193: does it help numerically to split up self.r2 and self.frequency**2? why not just keep them together as k2?

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+            argument
+        ) + self.config.amplitude_contrast * torch.cos(argument)
+
+        if self.config.bfactor == 0:
+            decay = (
+                np.sqrt(-np.log(self.config.valueNyquist))
+                * 2.0
+                * self.config.pixel_size
+            )
+            envelope = torch.exp(-self.frequency ** 2 * decay ** 2 * self.r2)
```
geoffwoollard, line 241: can you put in some documentation as to why this is a good default choice? where did it come from?

geoffwoollard: like what do we need (prevent aliasing? needs to go to zero at Nyquist or not? etc)

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+
+        self.config = config
+        self.wavelength = self._get_ewavelength()
+        self.frequency = 1.0 / (self.config.ctf_size * self.config.pixel_size)
+        n2 = float(self.config.ctf_size // 2)
+        ax = torch.arange(-n2, n2 + self.config.ctf_size % 2)
+        mx, my = torch.meshgrid(ax, ax)
+        self.register_buffer("r2", mx ** 2 + my ** 2)
+        self.register_buffer("r", torch.sqrt(self.r2))
+        self.register_buffer("angleFrequency", torch.atan2(my, mx))
```
geoffwoollard, line 178: angle_frequency

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        self.frequency = 1.0 / (self.config.sidelen * self.config.pixel_size)
+
+        n2 = float(self.config.sidelen // 2)
+        ax = torch.arange(-n2, n2 + self.config.sidelen % 2)
+        ax = torch.flip(ax, dims=[0])
+        mx, my = torch.meshgrid(ax, ax)
+
+        # shape SizexSize
+        self.register_buffer("mx", mx.clone())
+        self.register_buffer("my", my.clone())
```
geoffwoollard, line 301: what does this register_buffer do?

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+import numpy as np
+import starfile
+import torch
+import torch.fft
+from pytorch3d.transforms import (
+    euler_angles_to_matrix,
+    matrix_to_euler_angles,
+    random_rotations,
+)
+
```
geoffwoollard, line 13: looks like really good stuff!
https://youtu.be/Pph1r-x9nyY

geoffwoollard: are you sure you can install it through pip OK? https://youtu.be/Pph1r-x9nyY?t=96

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+    Returns
+    -------
+    config: class
+    """
+    config.starfile_available = config.input_starfile_path != ""
+    if config.starfile_available:
+        config = starfile_opticsparams(config)
+    config.ctf_size = config.sidelen  # TODO: make it adaptable
+    print(
```
geoffwoollard, line 34: log

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+    """
+
+    def __init__(self, config):
+
+        self.counter = 0
+        self.invert_hand = False
+        self.df = starfile.read(config.input_starfile_path)
+        check_star_file(config.input_starfile_path)
+        self.config = config
+        print("Reading parameters from the input starfile.")
```
geoffwoollard, line 95: log

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+            "relion_AngleTilt": torch.Tensor
+                Tensor of size (chunks,) that contains the Tilt angle (in degrees)
+                 for ZYZ rotation
+            "relion_AngleRot": torch.Tensor
+                Tensor of size (chunks,) that contains the Rot angle (in degrees)
+                for ZYZ rotation
+        """
+        AnglePsi = np.array(self.particle["rlnAnglePsi"], ndmin=2).squeeze()
+        AngleTilt = np.array(self.particle["rlnAngleTilt"], ndmin=2).squeeze()
+        AngleRot = np.array(self.particle["rlnAngleRot"], ndmin=2).squeeze()
```
geoffwoollard, line 138: I would modularize things so that we read it a compSPI (non-relion) star file, and there's a separate text to text .star file conversion step.

You can use the same headings if you want, but just rename `rlnAnglePsi` to `AnglePsi` or something

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+        relion_euler_np = np.radians(
+            np.stack(
+                [
+                    -AnglePsi,  # convert Relion to our convention
+                    AngleTilt * (-1 if self.config.relion_invert_hand else 1),
+                    # convert Relion to our convention + invert hand
+                    AngleRot,
+                ]
+            )
+        )  # convert Relion to our convention
```
geoffwoollard, line 148: I think this should be modularized outside, so that there's no upstream program specific conversions done in the code, but all in one place upstream.

simSPI/linear_simulator/params_utils.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @abstractstaticmethod
+    def get_params(self):
+        """Get the parameters."""
+
+
+"""Module to generate parameters using the input star file."""
+
+
+class starfile_params(Iparams):
```
geoffwoollard, line 78: for this class, I would modularize things now to have options for relion, cryosparc, etc

You don't have to implement parsing input data, but I would just do the software design so that we can go directly from external programs to compSPI, instead of something like cryosparc -> relion -> compSPI

simSPI/linear_simulator/utils.py
```diff
@@ -0,0 +1,136 @@
+"""contain miscellaneous function."""
+import numpy as np
+import torch
+
+
+def U2T(x):
```
geoffwoollard, line 6: uniform_to_triangular

simSPI/linear_simulator/starfile_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            ]
+
+        list_var += [
+            config.kV,
+            config.pixel_size,
+            config.cs,
+            config.amplitude_contrast,
+            config.bfactor,
+        ]
+        dataframe.append(list_var)
```
geoffwoollard, line 128: since this is not a dataframe, but a list, perhaps name it something that hints at that

simSPI/linear_simulator/test_data_creator.py
```diff
@@ -0,0 +1,86 @@
+
```
geoffwoollard, line 1: should this be in tests?

simSPI/tests_simple_simulator/test_ctf.py
```diff
@@ -0,0 +1,18 @@
+"""Test function for CTF module."""
+from ..linear_simulator.modules import CTF
+from ..linear_simulator.utils import init_data, normalized_mse, primal_to_fourier_2D
+
+
+def test_ctf():
+    """Test accuracy of the ctf."""
+    path = "simSPI/tests_simple_simulator/tests_data.npy"
```
geoffwoollard, line 8: shouldn't you re-generate this data? if the ctf code is changed, and things are consistent, but the new output doesn't correspond to `tests_data.npy`, wouldn't this test fail?

geoffwoollard: same goes for other tests using this .npy file (Noise, Projector, Shift, etc)

geoffwoollard (changes requested):
The tests weren't passing when I did this PR. 

Let's see if your tests pass.

ninamiolane (changes requested):
Thanks @harshit-gupta-cor for this amazing work! LOTS of code, very impressive. Congrats on writing so much clean code, and associated unit-tests 🙌  I really like the organization of the operators (shift, projector, ctf) as torch modules.

Sadly, it is impossible to review so much code at once (except for  @geoffwoollard maybe, who is superhuman). Could you split this PR into several smaller PRs, that we can tackle one by one?

Meanwhile, I have high-levels comments:

(1) *Consistency among simulators:* @geoffwoollard @harshit-gupta-cor could you make sure that the structure of both simulators are similar? Ideally, we would create an abstract Simulator class, that would serve as an blueprint to force each of its child class to implement the same methods:
https://www.geeksforgeeks.org/abstract-classes-in-python/

Copy-pasted from the above link:
An abstract class can be considered as a blueprint for other classes. It allows you to create a set of methods that must be created within any child classes built from the abstract class.

(2) *File organization and user experience:* Additionally, I see that this PR offers to create a `linear_simulator` folder, which asks the question on how we organize several simulators in our package. Do we want:
1. one folder per simulator: `/linear_simulator/` and `/multislice_simulator/`
or
2. one folder `/simulators/` with files
- `__init.py__`
- `base.py`: abstract class for the simulator `Simulator`
- `tem.py`: class for the TEM simulator
- `linear.py`: class for linear simulator
Option 1 means that the user writes: `from simSPI.linear_simulator import LinearSimulator`
Option 2 means that user writes: `from simSPI.simulators import LinearSimulator`
While Option 2 seems more natural from a user's perspective, maybe the structure of the code makes it easier to go for Option 1. 

(3) *Code style:* while the code is very clean, parts of it do not respect Python international conventions: see PEP8 https://www.python.org/dev/peps/pep-0008/.
These conventions exist so that developers code in the same style, and therefore understand each others' code rapidly. We should respect them if we want this project to scale (and we do 🦾 ).

(4) *Tests*: Filenames in tests should respect file names in the core code base. For example, if we want to test the folder `linear_simulator` there should be a file `test_linear_simulator`. And *not* `test_simple_simulator`. We need to understand in a fraction of a second which file is being tested.

(5) @harshit-gupta-cor did you make sure that some code that you are adding is not already in the library? I see computations around the ctf, and I remember functions having ctf computations too.

Excellent work, let's talk about it through the smaller PRs.

fredericpoitevin:
> Thanks @harshit-gupta-cor for this amazing work! LOTS of code, very impressive. Congrats on writing so much clean code, and associated unit-tests 🙌 I really like the organization of the operators (shift, projector, ctf) as torch modules.

This makes me think that the list of all historical contributors to this code should be acknowledged one way or another; Harshit, where relevant, would it be possible to add the name of Youssef and Julien as contributors in the docstrings? Did I forget others?

> 
> Sadly, it is impossible to review so much code at once (except for @geoffwoollard maybe, who is superhuman). Could you split this PR into several smaller PRs, that we can tackle one by one?

I am not sure it is easy to break this down in smaller bits unfortunately... 
But maybe the following requirements would be enough for this PR to be merged:
- the tests should all be passing
- we need to be convinced in some way that the simulator does a good job at simulating
- no major rehaul of the current structure of the repo (see Nina's remark on blueprint)

If that is not satisfactory, I'd suggest we discuss with Harshit about the plan on breaking down this PR in a series of smaller ones. The order might not be trivial to find.

README.md
```diff
@@ -11,11 +11,95 @@ Methods and tools for simulating SPI data.
 First create a conda environment with the required dependencies using the `enviroment.yml` file as follows:
 
     conda env create --file environment.yml
+or
+
+    conda env create -f environment.yml
+
```
fredericpoitevin, line 17: why this level of detail?

environment.yml
fredericpoitevin: we discussed it this morning, relaxing the version requirement might help?

simSPI/linear_simulator/configs/betagal_distribution.cfg
```diff
@@ -0,0 +1,71 @@
...
+            #str: name of the dataset folder
+    usegpu= False
+            #bool: (default: False) option to use GPU TODO: gpu option to be added
+    datasetSize= 100
+            #int: (default: 410) size of the dataset
+    chunks=10
+            #int: (default: 10) size of the chunk per iteration
+
+[paths]
+    input_volume_path = simSPI/linear_simulator/configs/betagal.mrc
```
fredericpoitevin, line 12: Unless they are used for tests, I am not sure we want to store data in a GitHub repo. The .mrc and .star files should be found somewhere else (see OSF - @geoffwoollard has been looking into that with his students); this would also be an opportunity to describe how these files were generated, where they come from, etc.

simSPI/linear_simulator/configs/betagal_distribution.cfg
fredericpoitevin: good point

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
+"""contains simulator and its components classes."""
```
fredericpoitevin, line 1: Replace with what you wrote below: "Module to generate data using a linear forward model."

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+
+import mrcfile
+import numpy as np
+import torch
+import torch.fft
+from pytorch3d.transforms import Rotate
+
+from .utils import fourier_to_primal_2D, init_cube, primal_to_fourier_2D
+
+"""Module to generate data using using liner forward model."""
```
fredericpoitevin, line 12: Remove

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+class simulator(torch.nn.Module):
+    """Class to generate data using liner forward model.
```
fredericpoitevin, line 16: linear

simSPI/linear_simulator/modules.py
fredericpoitevin: I'm not sure the order matters for speed (but I'm not an expert), though it does for broadcasting, which is also the reason why the second dimension exists here: https://pytorch.org/docs/stable/notes/broadcasting.html
To avoid confusion, I agree that it'd be nice to have a few words in the docstring on why this second dimension is needed for later.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+            )
+            volume = init_cube(self.config.sidelen)
+        else:
+            with mrcfile.open(self.config.input_volume_path, "r") as m:
+                volume = torch.from_numpy(m.data.copy()).to(self.Projector.vol.device)
+
+        self.Projector.vol = volume
+
+
+"""Class to generate tomographic projection."""
```
fredericpoitevin, line 87: Redundant with docstring below.

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        projection = torch.sum(
+            rot_vol.reshape(
+                batch_sz, self.config.sidelen, self.config.sidelen, self.config.sidelen
+            ),
+            dim=proj_axis,
+        )
+        projection = projection[
+            :, None, :, :
+        ]  # add a dummy channel (for consistency w/ img fmt)
+        # --> B,C,H,W
```
fredericpoitevin, line 150: This should be explained in the docstring instead (cf previous comment).

simSPI/linear_simulator/modules.py
fredericpoitevin: Correct, should be `-sqrt(1-a**2).sinx + a.cosx`

simSPI/linear_simulator/modules.py
fredericpoitevin: Good questions @geoffwoollard , I suggest someone opens an issue to discuss CTF and get to the bottom of this. There are several possible strategies and they all need to be explained and properly argumented.

simSPI/linear_simulator/modules.py
fredericpoitevin: Julien told me I could think of it as a way to send stuff to the GPU - he gave more details but I can only remember that part.
For more info, maybe check out: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.register_buffer

simSPI/linear_simulator/modules.py
```diff
@@ -0,0 +1,397 @@
...
+        n2 = float(self.config.sidelen // 2)
+        ax = torch.arange(-n2, n2 + self.config.sidelen % 2)
+        ax = torch.flip(ax, dims=[0])
+        mx, my = torch.meshgrid(ax, ax)
+
+        # shape SizexSize
+        self.register_buffer("mx", mx.clone())
+        self.register_buffer("my", my.clone())
+
+    def modulate(self, x_fourier, t_x, t_y):
```
fredericpoitevin, line 303: maybe `phase_shift` is a better name?

simSPI/linear_simulator/params_utils.py
fredericpoitevin: Maybe worth an issue to discuss?

simSPI/linear_simulator/starfile_utils.py
```diff
@@ -0,0 +1,129 @@
+"""contains the function to load parameters from the starfile."""
```
fredericpoitevin, line 1: maybe belongs in ioSPI?

simSPI/linear_simulator/utils.py
```diff
@@ -0,0 +1,136 @@
+"""contain miscellaneous function."""
```
fredericpoitevin, line 1: This file should not be merged. The content needs to be split in places where it belongs.

simSPI/tests_simple_simulator/test_ctf.py
fredericpoitevin: One could argue this would allow to detect when the CTF code changes and act accordingly.

fredericpoitevin (changes requested):
I agree with @ninamiolane that this is way more code to review than reasonable - my bad! I suggest we meet to discuss a plan for splitting it in a series of smaller PRs. Harshit, could you see if we could organize such a meeting? And would you agree to try to come up with a plan for the split that you could present then?

simSPI/linear_simulator/modules.py
harshit-gupta-cor: @geoffwoollard Great point! The CTF is negative at 0 in most of the literature. That is why we had the negative sign in CryoGAN (see [here](https://github.com/harshit-gupta-cor/CryoGAN/blob/ba4d33471668e79ccdd0a7706ba04d7d90a4465c/Functions/FunctionsCTF.py#L110))

However, I have tried to emulate the Relion's simulator and it uses the positive sign. The code in this PR (from cryonettorch) has been thoroughly compared w.r.t Relion and is consistent with it. Below is the CTF that is outputted using relion_project function. I am attaching a directory with .py and .sh files to replicate this image. It obtains projection from relion for a cube which is one pixel wide (delta volume). The fourier transform of the output is actually the CTF multiplied by a fourier normalization factor.
image

harshit-gupta-cor: @fredericpoitevin I think you meant: -sqrt(1-a**2).sinx - a.cosx 
Because then at x=0 value of CTF will be -a.cos(0)=-a=negative.

This is exactly what we had in CryoGAN and I have described above the reason why it was changed.

simSPI/linear_simulator/modules.py
harshit-gupta-cor: @geoffwoollard  Not sure what you mean by default choice. This piece of code sees if the bfactor of the envelope is 0 or not. If it is not zero then bfactor is used to create the envelope using the conventional formula. If it is zero then the value_nyquist is used which basically is given by user to say what is value of the envelope they want at nyquist frequency (see https://github.com/harshit-gupta-cor/simSPI/blob/004cbbda15fac64a721501cd9f41b5fe94553882/simSPI/linear_simulator/configs/betagal_distribution.cfg#L40). This is a more intuitive thing to give than bfactor. 

I agree this needs more documentation in the docstring which I will do. I don't think the value of bfactor determines aliasing. The latter I believe depends on the number of the samples we take of CTF between 0 and nyquist frequency (size of the CTF).

simSPI/linear_simulator/params_utils.py
harshit-gupta-cor: Yes, I installed it using pip. It has been put in the pip dependency in the new environement.yml and I was able to create and activate that environment.

simSPI/linear_simulator/params_utils.py
harshit-gupta-cor: Yeah, we had an extensive discussion on this with @fredericpoitevin  and @ynashed. We concluded that we will keep a [factory method pattern](https://www.geeksforgeeks.org/factory-method-python-design-patterns/) for the [params_generator](https://github.com/harshit-gupta-cor/simSPI/blob/0a3601c44fc9e8a72e058dfba74f22231e316398/simSPI/linear_simulator/dataset_generator.py#L34) class. Basically the rest of the code is blind to the params_generator and only tells it where it wants the params to come from (relion, distribution, or cryosparc in future). Each of the different options are put in the [factory](https://github.com/harshit-gupta-cor/simSPI/blob/0a3601c44fc9e8a72e058dfba74f22231e316398/simSPI/linear_simulator/params_utils.py#L44).

simSPI/linear_simulator/test_data_creator.py
harshit-gupta-cor: Removed this file. Actually, it was just a script to create the data for tests. Not needed in the repo.

harshit-gupta-cor:
@ninamiolane  @fredericpoitevin Thanks for looking into it. Yes, I agree smaller PRs will make it efficient. I think that is doable. I will send an email to set the meeting.

README.md
geoffwoollard: I would write `then shifted in the Fourier domain using scpecified shift values.`

example_scripts/generate_betagal_data_distributional.sh
geoffwoollard: you can just put what this bash script does in a readme, or in the documentation somehow

simSPI/linear_simulator/modules.py
geoffwoollard: https://agilescientific.com/blog/2018/12/28/what-is-the-fastest-axis-of-an-array

simSPI/linear_simulator/modules.py
geoffwoollard: https://github.com/compSPI/simSPI/issues/21

simSPI/linear_simulator/params_utils.py
geoffwoollard: I put in a comment in the issue on simulator inputs and outputs. We can continue on with the discussion in the issue, even after the PR is merged https://github.com/compSPI/simSPI/issues/14

geoffwoollard:
> * While Option 2 seems more natural from a user's perspective, maybe the structure of the code makes it easier to go for Option 1.

I like option 1 because it keeps the files for each simulator tidily nested in their folder.

geoffwoollard:
> (5) @harshit-gupta-cor did you make sure that some code that you are adding is not already in the library? I see computations around the ctf, and I remember functions having ctf computations too.

There's some numpy simulator code in compSPI, but rather than port things from there, I wanted to wait to see things here.

That simulator got projections by taking slices in Fourier space. I would like to implement that in this simulator, taking care to keep things autodiff friendly. 

In general
* When there are variations around a core simulator method, for instance the linear_simulator with some slights extensions, it might make sense to use exactly the same ctf. 
* but there are many ways to compute the ctf, and it can cause headaches to try to always have things tying back to the same ctf code, unless the simulator methods are very closely related. Of course, when possible, let's go for modularity. For instance it should be perfectly possible to extend the CTF in the linear_simulator to have astigmatism. 
* But what about a CTF with zernike polynomials (relion does this)? It might make sense to just use the another function that computes the ctf in a different way. Let's cross that bridge when we get there.

simSPI/linear_simulator/dataset_generator.py
```diff
@@ -0,0 +1,139 @@
...
+from .modules import simulator
+from .params_utils import ParamsFactory
+from .starfile_utils import return_names, starfile_data
+
+"""Module to generate and save dataset (including metadata) in the output directory."""
+
+
+class dataset_gen(torch.nn.Module):
+    """class to generate and save dataset (including metadata) in the output directory.
```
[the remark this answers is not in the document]
geoffwoollard: once the dust clears with the PRs and tests on the tem_simulator wrapper, we can design a way upstream of the tem_simulator and linear_simulator to have them organized the same way.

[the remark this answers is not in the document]
geoffwoollard: Thanks for sending that link, which I read in detail. I will try to get up to speed with all the lingo (child classes, abstract class, abstract method, declaration, implementation, subclassing, register the class explicitly). If you have any more, I'm all ears. 
 
I’ve gone through good ol’ Corey Schafer’s videos on object oriented programming… but that very basic
https://youtu.be/ZDa-Z5JzLYM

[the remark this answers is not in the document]
geoffwoollard: Let's wait for the dust to settle in the tem_simulator PRs, and then design something upstream of that and the linear_simulator so they have similar organization.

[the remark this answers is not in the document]
geoffwoollard: > Maybe we could have them inherit from an abstract Simulator class, that would itself inherits from torch.nn.Module and abc.ABC? See: https://www.geeksforgeeks.org/abstract-classes-in-python/

I see your point here. Let's continue discussion in https://github.com/compSPI/simSPI/issues/14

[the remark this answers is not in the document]
geoffwoollard: https://github.com/compSPI/simSPI/issues/14#issuecomment-983123756

fredericpoitevin:
> * When there are variations around a core simulator method, for instance the linear_simulator with some slights extensions, it might make sense to use exactly the same ctf.
> * but there are many ways to compute the ctf, and it can cause headaches to try to always have things tying back to the same ctf code, unless the simulator methods are very closely related. Of course, when possible, let's go for modularity. For instance it should be perfectly possible to extend the CTF in the linear_simulator to have astigmatism.
> * But what about a CTF with zernike polynomials (relion does this)? It might make sense to just use the another function that computes the ctf in a different way. Let's cross that bridge when we get there.

Ideally, we would like the CTF to be one object that can take many forms, depending on the level of accuracy (Zernike) or approximation (dampening versus padding). Julien taught me that this could be done using the concept of https://en.wikipedia.org/wiki/Polymorphism_(computer_science).
In practice, you could have many classes (possibly inheriting from a single trivial class, in the spirit of Nina's point about abstract classes) that are structured similarly, making their instantiation equivalent and just a matter of choice for the task at hand.

fredericpoitevin:
### Plan for series of PRs:

Individual branches from main will be created with a couple files only; some branches will need to be created sequentially, some in parallel. Total: 7 PR and 2 issues (roughly).

1. `modules.py`: split into 5 parallel PRs
1.1 `ctf_modules.py`: one PR in simSPI with unit tests
1.2 `noise_modules.py`: one PR in simSPI with unit tests
1.3 `shift_modules.py`: one PR in simSPI with unit tests
1.4 `projector_modules.py`: one PR in simSPI with unit tests
1.5 `simulator_modules.py`: one PR in simSPI with unit tests - need to discuss (see below).
2. `starfile_utils.py` and `dataset_generator.py`: this will go in a separate PR in ioSPI with unit tests, then a PR in simSPI for pytorch wrapping.
3. `params_utils.py`: this will go in a separate PR in simSPI.
4. `main.py` and `config.py`: one PR after the other PR are merged in simSPI, with unit tests and the `README.md` that already exists.
5. `utils.py`: this file disappears and a separate issue will be created to discuss where things go.
6. Discussion on how to integrate different simulators: create an issue in simSPI for that.

ninamiolane:
Excellent!! 💯  Thanks for making the effort of splitting the PR! @harshit-gupta-cor @ynashed @fredericpoitevin  

It will be worth it, see for example their numerous benefits from the official Google guidelines that strongly advocate for small PRs (also called CLs, changelist):

https://google.github.io/eng-practices/review/developer/small-cls.html

simSPI/linear_simulator/modules.py
harshit-gupta-cor: Agreed. Changed in the new [PR](https://github.com/compSPI/simSPI/pull/25).

simSPI/linear_simulator/modules.py
harshit-gupta-cor: Thanks! Changed in the new [PR](https://github.com/compSPI/simSPI/pull/23).

simSPI/linear_simulator/modules.py
harshit-gupta-cor: Done. Changed in the new [PR](https://github.com/compSPI/simSPI/pull/23/).

simSPI/linear_simulator/modules.py
harshit-gupta-cor: This is actually pytorch convention of keeping images in batch x channel x height x width format. This is important for example for the CNNs to understand which dimensions are batch, channel, etc. I have added a line in the docstring of this method in the new [PR](https://github.com/compSPI/simSPI/pull/23/).

simSPI/linear_simulator/modules.py
harshit-gupta-cor: What you are saying is quite relevant for reconstruction and data generation. However, this module gives the exact projection of a volume. Masking or any other processing should happen before/after this module.

simSPI/linear_simulator/modules.py
harshit-gupta-cor: You are right. Thanks! These names have been/will be changed according to the python convention in the new PRs.

simSPI/linear_simulator/configs/betagal_distribution.cfg
harshit-gupta-cor: I prefer cs notation. Its more readable. I added that its in mm in the docstring of CTF. Done in #27.

simSPI/linear_simulator/configs/betagal_distribution.cfg
harshit-gupta-cor: I prefer variable names without units. Its more readable. The units can be given in the docstring. For example, the docstring of CTF. Done in #27.

simSPI/tests_simple_simulator/test_ctf.py
harshit-gupta-cor: I agree with @fredericpoitevin It is much more secure for future.

fredericpoitevin:
Most of the plan was executed. Points 4. and 6. (see #28) remain, which constitute the refactoring step, so can be considered a separate task. Closing this PR now.

PR #25, opened 2021-12-06
Shift module

harshit-gupta-cor:
Splitted shift module out of modules.py as per the [discussion](https://github.com/compSPI/simSPI/pull/20#issuecomment-983853121). The comments from the previous [PR](https://github.com/compSPI/simSPI/pull/20) have been incorporated including following python name conventions.


harshit-gupta-cor:
Okay to review it!
Deepsource has a problem with the default argument {} for shift_params in shift_utils.py. 
Using {} as the default argument is quite fundamental to this code and I would prefer to keep it. 
Therefore, I think this issue can ignored for now and can be discussed more as the comments of this PR.

simSPI/linear_simulator/shift_utils.py
```diff
@@ -0,0 +1,90 @@
+"""Module containing method to shift an image in Fourier domain."""
```
fredericpoitevin, line 1: NIT: Module containing a method to apply a spatial shift on an image in Fourier domain.

simSPI/linear_simulator/shift_utils.py
```diff
@@ -0,0 +1,90 @@
...
+import numpy as np
+import torch
+
+
+class Shift(torch.nn.Module):
+    """class containing method to shift an image in Fourier domain.
```
fredericpoitevin, line 7: NIT: class containing a method to apply a spatial shift on an image in Fourier domain.

simSPI/linear_simulator/shift_utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+    """
+
+    def __init__(self, config):
+        super(Shift, self).__init__()
+
+        self.config = config
+        self.frequency = 1.0 / (self.config.side_len * self.config.pixel_size)
+
+        n2 = float(self.config.side_len // 2)
```
fredericpoitevin, line 24: I think `n_half_len` would be more explicit. `n2` could suggest nsquared for example.

simSPI/linear_simulator/shift_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Written by H. Gupta and Y.S.G. Nashed.
+
+    Parameters
+    ----------
+    config: class
+        Class containing parameters of the shift
+
+    """
+
+    def __init__(self, config):
```
fredericpoitevin, line 18: add a docstring that states what is being done here. Something like 
```
""" Initialize image grid. """
```

tests/test_shift.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        config_dict = {}
+    config = AttrDict(config_dict)
+    return saved_data, config
+
+
+def normalized_mse(a, b):
+    """Return mean square error.
+
+    Calclulates mean square error between two inputs normalized
```
fredericpoitevin, line 49: NIT: Calculates

simSPI/linear_simulator/shift_utils.py
```diff
@@ -0,0 +1,90 @@
...
+        t_y = t_y[:, None, None, None]
+        t_x = t_x[:, None, None, None]
+
+        modulation = torch.exp(
+            -2 * np.pi * 1j * self.frequency * (self.mx * t_y + self.my * t_x)
+        )
+
+        return x_fourier * modulation
+
+    def forward(self, x_fourier, shift_params={}):
```
fredericpoitevin, line 62: Why would it not be possible to do what [Deepsource suggests to do](https://deepsource.io/gh/compSPI/simSPI/run/e9234511-4b7)?

My understanding is that it would look something like this:
```python
    def forward(self, x_fourier, shift_params=None):
        if shift_params is not None:
           x_fourier = self.phase_shift(
               x_fourier, shift_params["shift_x"], shift_params["shift_y"]
            )
        return x_fourier
```

fredericpoitevin (commented):
Looking great @harshit-gupta-cor !

Please let's discuss below the couple comments I've made before merging.

Thanks,
Fred

harshit-gupta-cor:
Thanks @fredericpoitevin. I have incorporated all the suggestions including the None default. The default to {} was done just to keep it like a dict but I think the None default is manageable too.

The updated PR passes all the test except a test which is from TEM simulator and nothing to do with this PR. @geoffwoollard you might wanna look into it. The test function is written for a function which calls a random variable inside it. This is making it pass most of the times but fail a few times.

fredericpoitevin (approved):
Thank you @harshit-gupta-cor ! Sorry I could not approve earlier, great job!

PR #39, opened 2021-12-13
Revert "Feature: adding helper functions from SLAC"

arjunsingh3600:
Reverts compSPI/simSPI#19


fredericpoitevin:
Merging this so the main branch reverts to a safe state. @arjunsingh3600 you can go ahead and create a new PR with that will replicate #19 so we can fix it properly.

PR #32, opened 2021-12-10
Tem Implementation with tests

arjunsingh3600:
This is the TEM wrapper implementation without the run() and get_image_data() functions (they require a tem sim installation to run).
This PR is blocked by -

- https://github.com/compSPI/ioSPI/pull/24
- https://github.com/compSPI/simSPI/pull/19

@jedyeo @geoffwoollard @thisTyler

PR #42, opened 2021-12-14
Add tem wrapper helpers

arjunsingh3600:
This branch contains all the changes from #39. The history of this branch is recorded there. This new PR serves to fix any tests that prevent a merge.


arjunsingh3600:
@jedyeo @fredericpoitevin @geoffwoollard This branch should be ready to merge.

simSPI/crd.py
```diff
@@ -0,0 +1,78 @@
...
+    """Return a rotation list containing Euler angles.
+
+    Parameters
+    ----------
+    n_particles : int
+        Number of particles.
+    """
+    rotlist = []
+    for _ in range(n_particles + 1):
+        x = R.random().as_euler("zyz", degrees=True)
```
geoffwoollard, line 76: I think it's better just to do `from scipy.spatial.transform import Rotation` and not use `R` as shorthand. It makes it easier to search and replace. No need to be super compact.

simSPI/crd.py
```diff
@@ -0,0 +1,78 @@
...
+
+Write particle stack data (slice of protein)
+to .txt file containing coordinate (crd) information on particle.
+"""
+
+import logging
+import os
+
+import numpy as np
+from scipy.spatial.transform import Rotation as R
```
geoffwoollard, line 11: https://github.com/compSPI/simSPI/pull/42/files#r768833474

tests/test_tem.py
```diff
@@ -1,11 +1,11 @@
...
-import tempfile
 import os
+import tempfile
 
 import yaml
 
+from simSPI import tem
+
 
 def test_tem_main():
```
geoffwoollard, line 10: are these tests running tem simulator on github? how since I thought there needed to be a local install? 

if the tests are in a try catch, I think that's a good fix for now, but you could just put a note in the documentation about that. something like "sim.run in try catch because requires local install and currently no containerization is set up")

geoffwoollard (approved):
a few minor things to address, but can be merged after that.

let's keep the main branch always clean since people are branching off of it and all the tests need to pass!

simSPI/crd.py
fredericpoitevin: I agree

tests/test_tem.py
fredericpoitevin: I agree, it would help to know what is being tested.

fredericpoitevin (approved):
Looking good! Please address Geoff's points and then we can merge :D

tests/test_tem.py
arjunsingh3600: These changes seem to be from one of the merges. This test_tem.py is essentially the same as the one in the main branch right now. It was put in place to provide code coverage for the _function definitions_ not the implementation of the tem_wrapper, hence it passes. The actual tem_implementation with associated tests will be merged with #32.

geoffwoollard: ok let's merge for now. but I stand by my point

PR #45, opened 2021-12-15
solve pytorch3d dependency issue - platform dependence

harshit-gupta-cor:


fredericpoitevin:
@harshit-gupta-cor I don't think this is the right fix. We need to have install instructions separate for Mac and Linux as detailed here: https://github.com/facebookresearch/pytorch3d/blob/main/INSTALL.md#installing-prebuilt-binaries-for-pytorch3d

environment.yml
```diff
@@ -19,6 +20,7 @@ dependencies:
   - pytest
   - pytest-cov
   - pytorch=1.9.0
+  - pytorch3d
```
ninamiolane, line 23: Do we need pytorch3d twice?

ninamiolane:
@harshit-gupta-cor @fredericpoitevin what is the update on this? is this the right fix or not?

harshit-gupta-cor:
> @harshit-gupta-cor @fredericpoitevin what is the update on this? is this the right fix or not?

 Sorry my bad for not updating this. The commit was done so that the environment works for both mac and linux by having pytorch3d installation from both conda (for linux) and pip (for  mac). However, we decided that we will keep only linux version.

Therefore, I will just remove pytorch3d from pip. I will commit that ASAP. Since I have mac, I will first check if the changed environment is installable on sdf cluster and then will commit.

ninamiolane:
@harshit-gupta-cor fantastic, thanks!

harshit-gupta-cor:
@ninamiolane Changes done. This can be merged now.

ninamiolane (approved):
Great, thanks @harshit-gupta-cor !!

PR #51, opened 2022-01-24
Decoupled linting and testing, and updated badges in README.

fredericpoitevin:


ninamiolane (approved):
LGTM! 💯 Two questions:
- the python version is not showing on the test workflow (whereas it is showing on the lint workflow): any chance we could add it? It would help knowing which python version the github actions are running, if it fails
- is it normal that we still mention "singularity"?

fredericpoitevin:
@ninamiolane what we could look into is the following: 
- have `build_container.yml` build one image per "strategy" (ubuntu version, python version), with the image tag reflecting the version
- have `action.yml` somehow taking the tag as argument.
I'll look into that.

PR #62, opened 2022-01-29
Still trying to fix test_tem.py

fredericpoitevin:


fredericpoitevin:
Linting test is not passing anymore, which does not make sense since it passed in the version of `master` this branched out from and none of the two files `blake8` is complaining about has been changed:
```
Run black . --check
  black . --check
  isort --profile black --check .
  shell: /bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.10/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.10/x64/lib
would reformat simSPI/transfer.py
would reformat tests/test_fov.py

Oh no! 💥 💔 💥
2 files would be reformatted, 15 files would be left unchanged.
Error: Process completed with exit code 1.
```
I'm going to merge and see what happens.

PR #63, opened 2022-01-30
Fix asynchronous testing in container.

fredericpoitevin:
Attempt at updating simspi:latest only if the current version of the code passes. This would prevent asynchronous tests as we had before.


fredericpoitevin:
@ninamiolane I have mixed feelings about the issue raised by DeepSource: should we ignore it? https://deepsource.io/gh/compSPI/simSPI/run/2f657bf0-d7cb-46cb-9dd2-5526fb32b948/python/PTC-W0028

geoffwoollard:
I've seen that error about using some in built hypotenuse vs `rho = np.sqrt(x**2 + y**2)`. I think it's not helpful and things are fine as they are. It's a nice suggestion, but wouldn't be an issue here since x and y coming out of `np.meshgrid` are well behaved.

ninamiolane:
Sounds good, let's just ignore this DeepSource issue! (I've just ignored it there) @fredericpoitevin @geoffwoollard

ninamiolane (approved):
Very nice!! Thank you so much @fredericpoitevin for getting to the end of this difficult task!

fredericpoitevin:
Thanks @ninamiolane for setting up the ignore in Deepsource.

@geoffwoollard I edited to `hypot` according to the recommendation, but it failed because it only accepts scalar - which deepsource did not realize. I agree we are probably safe with `np.sqrt(x**2 + y**2)`.

PR #67, opened 2022-02-12
Simulator volume utils

harshit-gupta-cor:
This PR follows PRs https://github.com/compSPI/simSPI/pull/20.
PR https://github.com/compSPI/simSPI/pull/20 asked for smaller PRs (see discussion).


fredericpoitevin:
I think the issue is the following: updating the environment requires updating the container, which fails from a fork other than compSPI because of DockerHub credentials only setup for that fork.
I created a simulator_volume_utils branch on the compSPI fork, will close the current PR and create a new one from there. The credentials should then work.

PR #71, opened 2022-02-13
Param utils

harshit-gupta-cor:
This PR follows the discussion in #20. The print statements will be replaced with loggers once the full pipeline is on the repo.

PR #74, opened 2022-02-16
Add utils to handle simulation parameters (including distributional parameters) in linear simulator: rotations, shifts, CTF

harshit-gupta-cor:
Replaces PR #71.


harshit-gupta-cor:
@fredericpoitevin to me this PR looks ok to merge. I resolved the reasonable issues raised by deepsource for PR #71. The current issues have been raised for [factory method design pattern] which I have followed from our discussions and standard materials on design patterns for ex. [here]((https://www.geeksforgeeks.org/factory-method-python-design-patterns/)).  There is one issue for test functions which I dont find to be relevant.

geoffwoollard (approved):

ninamiolane (approved):
LGTM, if @fredericpoitevin you confirm that we should ignore the github issues. Thanks!

PS: It is usually not useful to call a file _utils.py (it does not add any information, it just makes the file name longer); but I assume this will be refactored when both simulators are merged.

fredericpoitevin:
@ninamiolane I don't think we should merge this, the deepsource issues should be fixed first. I've started working on them.

fredericpoitevin:
OK I think I've fixed a few. I'm running out of steam with [this one](https://deepsource.io/gh/compSPI/simSPI/run/b7d37168-3045-466e-b5e4-618555ad1fc0/python/PYL-W0201) - which we typically solved with adding more config files to the test data, one per case considered. This will need to be taken care of eventually but I'll go ahead and merge so we can progress toward refactoring.

PR #79, opened 2022-02-23
Fix: appropriate slicing of stack

jedyeo:
Small change that allows for proper slicing of particle stack. When displaying the micrograph the previous code would stack the slices on top on each other while iterating through the rotations for subsequent slices. This change allows for proper iteration through the particle stacks.


geoffwoollard:
Tests failing 
```
simSPI/tem.py:9: in <module>
    from ioSPI.ioSPI import micrographs
E   ModuleNotFoundError: No module named 'ioSPI.ioSPI'
```

Should it be `from ioSPI import micrographs`?

fredericpoitevin:
@calhep is it OK to delete the `tem_metadata_v2` branch?

Issue #35, opened 2021-12-11
helper functions to sample ctf parameters (for tem simulator and other simulators)


geoffwoollard:
see simSPI.transfer.random_ctfs https://github.com/compSPI/simSPI/blob/master/simSPI/transfer.py#L109

in particular
```
    dfs = np.random.uniform(low=df_min, high=df_max, size=n_particles)
    df_diff = np.random.uniform(low=df_diff_min, high=df_diff_max, size=n_particles)
    df1s = dfs - df_diff / 2
    df2s = dfs + df_diff / 2
    df_ang_deg = np.random.uniform(low=df_ang_min, high=df_ang_max, size=n_particles)
```

For now, you are just passing in defocus (`- defocus [OPTIONAL]           : nominal defocus value in um`) so you could do something like
```
if defocus_distribution == 'uniform':
  dfs = np.random.uniform(low=defocus_min_ang, high=defocus_max_ang, size=n_particles)
elif defocus_distribution == 'gaussian':
  dfs = np.random.gaussian(loc=defocus_loc_ang, scale=defocus_scale_ang, size=n_particles)
```

This involves reading in some input meta data about the data generating process: what `defocus_distribution` is, and the params of that distribution: `defocus_min_ang` and `defocus_max_ang` if `uniform`; `defocus_loc_ang` and `defocus_scale_ang` if `gaussian`. Set things up to extend to arbitrary distributions (e.g. mixture of gaussians, etc).

geoffwoollard:
Suggested deadline: Jan 14

arjunsingh3600:
Discussion from Slack : The TEM simulator mentions the following CTF. Why are we overlaying another one on top of simulated particle?
![image](https://user-images.githubusercontent.com/50425729/150070145-f94d77a6-dfe4-4443-8d51-2e3482553fbb.png)

we just want to give the CTF params to the TEM simulator. ie the delta z. the k comes from the beam energy (300 kev) and the Cs also, but those are typically the same for every image.
so we just need to specify the distributions that these ctf params come from, and keep track of what actual values were sampled.
then pass these sampled values to the tem simulator, which will do the simulating

arjunsingh3600:
@geoffwoollard Could you please verify my understanding of the issue here : 
1) The goal here to randomly generate the defocus parameter delta z with one of the following distributions : 

- uniform
- gaussian
- mixture of gaussian

2) The other parameters of the CTF i.e Cs and k would remain the same for every image

[...]

PR #82, opened 2022-03-08
Tem/ctf params

arjunsingh3600:
Functionality for defocus simulation parameters, as well as the ability to generate multiple datasets concurrently with associated metadata files.


simSPI/tem.py
```diff
@@ -8,7 +8,7 @@
 import yaml
 from ioSPI import micrographs
 
-from simSPI import crd, fov, tem_inputs
+from simSPI import crd, fov, tem_distribution_utils, tem_inputs
```
geoffwoollard, line 11: Drop the `_utils`, as it's not that informative.

simSPI/tem_distribution_utils.py
```diff
@@ -0,0 +1,52 @@
+"""Utility file for generating and sampling distributions."""
```
geoffwoollard, line 1: I like this file!

geoffwoollard: I think it's a good starting point for general stochastic simulators, where the distributional parameters are given, and not every single value.

sim_config.yaml
```diff
@@ -35,5 +35,12 @@
     detector_q_efficiency: 0.5         # detector quantum efficiency
     mtf_params: [0,0,1,0,0]            # to be described. [0,0,1,0,0] is perfect detector.
 
+  geometry_parameters:
+    n_samples: 4                       # number of samples to be generated
+
+  ctf_parameters:
+    distribution_type: 'uniform'        # type of distribution, either 'gaussian' or 'uniform'
+    distribution_parameters: [0, 1]     # distribution parameters, uniform: (low, high), gaussian: (loc, scale)
```
geoffwoollard, line 43: maybe have
```
loc=0
scale=1
```

explicitly

tests/test_files/path_config.yaml
```diff
@@ -4,3 +4,4 @@
   mrc_keyword: '_randomrot'
   output_dir: '/work/tests/test_files/'
   local_sim_dir: '/work/TEM-simulator_1.3/src/TEM-simulator'
```
geoffwoollard, line 6: get docker working

tests/test_tem_distribution_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    assert generator.distribution is not None
+
+    generator = tem_distribution_utils.DistributionGenerator("gaussian", (0, 1))
+    assert generator.distribution is not None
+
+    with pytest.raises(NotImplementedError):
+        tem_distribution_utils.DistributionGenerator("invalid_value", (0, 1))
+
+
+def test_draw_samples_1d():
```
geoffwoollard, line 20: I would make these tests more specific. Test max or min of uniform. Test some expectations from many samples from Gaussian (mean near zero, variance near scale**2). There is "sample error" but you can use some tolerance.

tests/test_tem_inputs.py
```diff
@@ -26,6 +28,7 @@ def test_fill_parameters_dictionary_max():
 
     mrc_file = "a.mrc"
     pdb_file = "a.pdb"
+    defocus_file = "a.txt"
     voxel_size = 0.2
```
geoffwoollard, line 32: should have units.

tests/test_tem_inputs.py
```diff
@@ -204,6 +216,7 @@ def test_fill_parameters_dictionary_min():
     detector_q_efficiency = 0.1
     mtf_params = [0.1, 0.0, 0.7, 0, 0]
     log_file = "itslog.log"
+    n_samples = 5
```
geoffwoollard, line 219: have n_particles as input. From particle size and size of micrograph, figure out how many micrographs needs to be simulated, then give run this many micrographs.

geoffwoollard (changes requested):
Good job so far. You're close to the finish line.

There are some issues that can be done in another PR
* docker not local tem simulator
* number of particles not number of micrographs

But I would request changes to two things in this PR
* documentation of tem_simulator. Readable sentences rather than [um], [V], etc
* yamls in test should be under tem_simulator or named with tem_simulator in their name so it's clear what they are for.

ninamiolane (approved):
LGTM as soon as you address Geoff's comments!

sim_config.yaml
arjunsingh3600: This change might be better in a future PR as it goes hand in hand with the comment above. I thought it'd be nice to brainstorm a structure that is a bit more verbose , so I added your comments to the issue here : https://github.com/compSPI/simSPI/issues/35

tests/test_tem_inputs.py
arjunsingh3600: The information for all these variables is in sim_config, To prevent duplication - I added a comment referring to the appropriate file for details on these variables.

arjunsingh3600:
@geoffwoollard Implemented the changes requested above.

tests/test_files/tem/path_config.yaml
```diff
@@ -0,0 +1,7 @@
+  # absolute paths
+  pdb_file: '/work/tests/test_files/tem/4v6x.pdb'
+  metadata_params_file: '/work/tests/test_files/tem/metadata_fields.yaml'
+  mrc_keyword: '_randomrot'
+  output_dir: '/work/tests/test_files/tem'
+  local_sim_dir: '/work/TEM-simulator_1.3/src/TEM-simulator'
```
geoffwoollard, line 6: are these local paths to your machine? where is `/work`?

geoffwoollard (approved):
Ok to merge if all the issues in the demo day were addressed. No local TEM simulator referenced, docker image instead, etc.

PR #86, opened 2022-03-10
feat: demo dataset gen notebook

jedyeo:


ninamiolane (changes requested):
There should be a test_notebooks.py file to test the notebook (or a shorter version of it, so that it does not take forever to run).

e.g. like this file:
https://github.com/geomstats/geomstats/blob/master/tests/test_notebooks.py

ninamiolane:
Also notebooks/.DS_Store should be in the gitignore

fredericpoitevin (changes requested):
Almost there! Thanks for taking into accounts most of our comments already. Please address Nina's comments and we're good to merge.

I'll create two issues regarding the points I raised:
- lack of logging during TEMsimulator.run()
- output formatting compatible with RELION

jedyeo:
This seems outdated, i.e. the notebook before refactoring. See #87 for demo notebook

PR #57, opened 2022-01-28
Adding usage tutorial and files related to it

luis-sribeiro:
First version of a tutorial to show some of the available functionalities.

PR #96, opened 2022-04-03
Luis master

fredericpoitevin:
Copy of #57 on main fork, for test.


fredericpoitevin:
Closes #57 since it is essentially the same with some fixed.

fredericpoitevin:
I opened #98 to address the fact that the "Building simSPI Container" check is skipped.

PR #100, opened 2022-04-07
Add __init__.py to linear_simulator directory

jedyeo:
`linear_simulator` needs to be recognized as a module so `reconstructSPI` can successfully import from this directory.

See https://github.com/compSPI/reconstructSPI/pull/48


jedyeo:
uh i dont know how that happened

